PR #3, opened 2016-04-05
Add unit test for signal handling

zzxuanyuan:
This is a unit test new signal handling. It tests four signals SIGBUS, SIGSEGV, SIGILL, SIGFPE. In each test case, there are three threads created and consistently allocating and deleting memory space.


xvallspl:
@pcanal Is this still relevant?

pcanal:
yes but waiting on me to update the relevant code in root.

etejedor:
@pcanal we still need to wait for the update of the corresponding code in ROOT for this PR?

PR #9, opened 2016-11-04
Add test for IMT of TTree::Flush.

bbockelm:
This is a simple test corresponding to PR 277 in the root mirror.


pcanal:
Merged.

Thanks,
Philippe.

PR #10, opened 2016-11-07
Extend TThreadExecutor tests with chunking

xvallspl:
Don't merge before merging pull request in ROOT's main repository: https://github.com/root-mirror/root/pull/289


lmoneta:
This is now committed in the master

PR #16, opened 2017-02-06
[TDF] add test_typeguessing.cxx

bluehood:
small test, only performs basic checks


dpiparo:
Hi,

would it be possible to change the name of the test root file in order to include the name of the test? This would avoid strange effects when running the tests in parallel (all other tests adopt this convention).

bluehood:
Sure, done!

dpiparo:
Merged, thanks.

PR #19, opened 2017-02-08
[TDF] sync tests with changes in ROOT

bluehood:
Also fixes test_branchoverwrite reference output


bluehood:
This patch is meant to be applied together with [these minor fixes](https://github.com/root-mirror/root/pull/336) in the ROOT main repo.

dpiparo:
Merged, thanks.

PR #23, opened 2017-02-14
[TDF] Remove redundant logic from tests

bluehood:
Now that all tests are executed with and without implicit mt activated, there is no need to test both cases separately (the upcoming `testIMT` will take care of that case explicitly).


dpiparo:
Merged, Thanks.

PR #25, opened 2017-02-20
[TDF] Add test_callables

bluehood:
This test verifies we support different callable types:
free functions, functor classes, lambdas and std::functions are used
as filters, temporary branch expressions and foreach expressions.
Usage of these callables is sprinkled with std::move's.


dpiparo:
Merged, thanks.

PR #26, opened 2017-02-20
[TDF] add test_reduce

bluehood:
`TDataFrame::Reduce` also supports branch types with no default constructor, but it seems that it was an overkill since it is not possible to have such objects in a `TTree`.


dpiparo:
Merged, thanks.

PR #27, opened 2017-02-20
[TDF] test_reduce: also check for support to objects with no def ctor

bluehood:


dpiparo:
Hi @bluehood , the PR seem to have fallen through the cracks: it's now merged.

PR #30, opened 2017-03-08
Label longtest for tests running over 1 minute

martinmine:
Second part of https://github.com/root-project/root/pull/408


peremato:
Merged

PR #33, opened 2017-03-30
[TSEQ] Add test for TSeq::iterator class

bluehood:


xvallspl:
Do we need reference files for tests like this? It already returns 0 if it passes. Seems redundant.

dpiparo:
no we do not actually. If no ref file is specified, we can check the retcode, by default expected to be 0.

bluehood:
My fault, sorry. I wouldn't open another PR just for that, maybe Danilo could you remove the reference with a ninja commit? :) Otherwise I'll open a PR

xvallspl:
Dit it myself, PR coming

dpiparo:
cool, looking forward!

bluehood:
thank you!

PR #42, opened 2017-05-16
Move call to EnableImpliclitMT() to the top of script test_emptysource.cxx

amadio:
The snapshot action in TDataFrame launches a separate
thread for merging data created by each slot, even when
only one slot is used. Therefore, either EnableThreadSafety()
or EnableImplicitMT() must be called beforehand.

### DIFF
--- a/root/dataframe/test_emptysource.cxx
+++ b/root/dataframe/test_emptysource.cxx
@@ -8,6 +8,11 @@
 #include <atomic>
 
 int main() {
+#ifdef R__USE_IMT
+   unsigned int numThreads = 4;
+   ROOT::EnableImplicitMT(numThreads);
+#endif
+
    auto fileName = "test_emptysource.root";
    auto treeName = "test_emptysource";
    Long64_t numEntries = 1002;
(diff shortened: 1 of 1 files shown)


dpiparo:
Hi @amadio ,
I would not impose this to the user. Can we have 2 codepaths , one with and one without imt enabled? We can have two lambdas or two composed lambdas perhaps...

amadio:
@dpiparo Sure, that makes sense. We could also just call EnableThreadSafety() inside the snapshot function. I will make a pull request with these changes. I wonder why the current master doesn't fail this test, as the same problem must exist.

dpiparo:
Hi @amadio,
It's a possibility, but I am a bit reluctant to change the state of ROOT with a method call of TDF though.

amadio:
Ok, so I should then just have two snapshot functions, one for single threaded, and one for MT mode, right?

dpiparo:
Hi, that is what I would go for if possible.

amadio:
Ok, I will close this then, as we will fix this elsewhere.

PR #43, opened 2017-05-16
Adapted tab completion options in ref files

xvallspl:
This PR goes together with https://github.com/root-project/root/pull/567


xvallspl:
I thought this PR was merged long ago! I guess it's not a problem anymore? Or did we stop checking for it?

PR #45, opened 2017-05-25
Update constant for invalid compression algorithm.

bbockelm:
Previously, `4` was used to indicate an invalid compression algorithm number.

Bump this in preparation for assigning compression algorithm `4` to LZ4.


root/io/compression/MainEvent.cxx
```diff
@@ -145,7 +145,7 @@ int main(int argc, char **argv)
    if (message->GetCompressionSettings() != 301) exit(153);
 
    message->SetCompressionSettings(202);
-   message->SetCompressionAlgorithm(4);
+   message->SetCompressionAlgorithm(5);
```
pcanal: rather than 5, can you make 99 or so :)

PR #46, opened 2017-06-02
misc fixes

bluehood:
* Proposed solution for `python-pythonizations-smartptr`: substitute usages of `auto_ptr` with `shared_ptr` to silence deprecation warnings. The test does not seem to rely on the semantic differences between the two.
* Revert a change that had been done to `root/dataframe/test_glob.C` in order to make failures more verbose. It is not needed and never really helped.

Feel free to close this PR without merging if you deem so.


python/pythonizations/SmartPtr.C
```diff
@@ -1,18 +1,18 @@
 #include "SmartPtr.h"
 
-auto_ptr<MyShareable> gime_mine() { return mine; }
-auto_ptr<MyShareable>* gime_mine_ptr() { return &mine; }
-auto_ptr<MyShareable>& gime_mine_ref() { return mine; }
+shared_ptr<MyShareable> gime_mine() { return mine; }
+shared_ptr<MyShareable>* gime_mine_ptr() { return &mine; }
```
pcanal: Why not use unique_ptr?

python/pythonizations/SmartPtr.h
```diff
@@ -22,25 +22,25 @@ class MyShareable {
...
-void pass_mine_sp_ptr(auto_ptr<MyShareable>* p);
+void pass_mine_sp(shared_ptr<MyShareable> p);
+void pass_mine_sp_ref(shared_ptr<MyShareable>& p);
+void pass_mine_sp_ptr(shared_ptr<MyShareable>* p);
 
 void pass_mine_rp(MyShareable);
 void pass_mine_rp_ref(const MyShareable&);
 void pass_mine_rp_ptr(const MyShareable*);
 
 #ifdef __GCCXML__
```
pcanal: This section can now be completely removed (i.e. __GCCXML__ is never defined)

python/pythonizations/SmartPtr.C
bluehood: It was my first choice but it cannot be returned by `gime_mine` as easily. Using a `shared_ptr` removes the warnings maintaining the semantics of the tests intact (as far as I can tell)

pcanal: What was the error message?

bluehood: A `unique_ptr` lvalue cannot be returned by value without moving, but moving the variable out of `gime_mine` is also not an option it is a global.

pcanal: right ... but it was also the prior behavior (implicit move in the auto_ptr copy constructor ...)

bluehood: As far as I can tell such behaviour is not relevant is this test (i.e. any smart pointer would do fine), and `shared_ptr` is the obvious choice for a global variable that is passed around and copied in and out of functions.

Of course I might be wrong, and it would be nice to be able to test this change before merging to master :smile: (on the other hand the revert would be super simple)

dpiparo: more importantly: why is this happening only on arm64 fst builds?

bluehood: good question: the warnings appear in all builds (e.g. http://cdash.cern.ch/testDetails.php?test=19878020&build=360727), arm actually crashes:

```
CMake Error at /data/sftnight/build/workspace/root-nightly-master/BUILDTYPE/Optimized/COMPILER/native/LABEL/arm64/root/cmake/modules/RootTestDriver.cmake:161 (message):
  error code: Child aborted
```

My commit should (but I can't test because the jenkins bot does not run on roottest PRs) resolve the warnings, which might or might not help with the failure on arm. Anyway, as I said in my first message feel free to close this if you don't agree on the approach.

pcanal: Reading the full test I agree with Enrico.

bluehood:
@pcanal @dpiparo ok now the `smartptr` test does not produce warnings but just fails :sweat: sorry!

If I understand correctly pyROOT has some logic to make an `auto_ptr` behave like its content type, but it does not do the same for a `shared_ptr`.

pcanal:
Ok, then let's go with unique_ptr ...

PR #61, opened 2017-07-18
[TDF] Speed up test_snapshotNFiles

bluehood:
...by compiling instead of jitting three Snapshot invocations.


peremato:
@phsft-bot build

PR #65, opened 2017-08-08
Add include directories for externals to include paths

amadio:
This is needed to avoid copying the headers of externals (Vc and VecCore) to `${CMAKE_BINARY_DIR}/include` which forces us to inherit all warnings from Vc.
The change introduced here adds the necessary includes when running roottest tests.
See, e.g., https://github.com/root-project/root/pull/830#issuecomment-320942852 for examples of failing tests due to the missing includes.


amadio:
Merging to test PR https://github.com/root-project/root/issues/830.

PR #70, opened 2017-08-29
[genreflex] Add tests for noStreamer and noInputOperator xml attribute

breitwieserCern:
What effect has `csr->SetRequestNoInputOperator(true);` on the dictionary generation?
I was expecting to find a `TBuffer &operator<<(TBuffer &buffer, const Foo *foo)` inside the dictionary if no input operator is set to false, but when I compared both options with:
```bash
diff -y roottest/root/meta/genreflex/noStreamer_noInputOperator/noInputOperator_rflx.cpp roottest/root/meta/genreflex/noStreamer_noInputOperator/noInputOperator_false_rflx.cpp
```
the only difference beside names was:
```
static ::ROOT::TGenericClassInfo 				      static ::ROOT::TGenericClassInfo 
         instance("Foo", ::Foo::Class_Version(), "", 9,		         instance("Foo", ::Foo::Class_Version(), "", 9,
                  typeid(::Foo), ::ROOT::Internal::DefineBeha	                  typeid(::Foo), ::ROOT::Internal::DefineBeha
                  &::Foo::Dictionary, isa_proxy, 6,	      |	                  &::Foo::Dictionary, isa_proxy, 4,
                  sizeof(::Foo) );				                  sizeof(::Foo) );
```

What am I missing?

### DIFF
--- a/root/meta/genreflex/noStreamer_noInputOperator/.rootrc
+++ b/root/meta/genreflex/noStreamer_noInputOperator/.rootrc
@@ -1,0 +1,1 @@
+Rint.History:            .root_hist
--- a/root/meta/genreflex/noStreamer_noInputOperator/CMakeLists.txt
+++ b/root/meta/genreflex/noStreamer_noInputOperator/CMakeLists.txt
@@ -1,0 +1,45 @@
+ROOTTEST_ADD_TESTDIRS()
+
+# ------------------------------------------------------------------------------
+# noStreamer tests
+ROOTTEST_ADD_TEST(invalidNoStreamerGenreflex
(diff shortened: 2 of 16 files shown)


dpiparo:
Nice. Could you refactor the CMakeList in order to take advantage of the functions to create dictionaries? Here the instructions: https://github.com/root-project/root/pull/864

dpiparo:
For the input operator, I need to check.

breitwieserCern:
The tests with invalid xml file still use ROOT_ADD_TEST to invoke genreflex since I expect the call to fail with a certain error message.

PR #71, opened 2017-08-30
Fix tests for root PR #831

Teemperor:
Rootcling PR [#831](https://github.com/root-project/root/pull/831) fixes some erros with the return codes and the verbosity level. However, some tests actually test the old broken behavior. This PR updates those tests that #831 doesn't break master.


Teemperor:
@phsft-bot build

PR #72, opened 2017-08-30
Filemerger test fixes

oshadura:
Investigation doesn't show any data differences in histograms, comparison shows that for TH1F/TH2F was added extra bytes, not affecting data of histograms. Hsimple101.root reference was updated according hsimple.root (with hadd -f1 option and default ZLIB algorithm)

### DIFF
--- a/root/io/filemerger/execTestMultiMerge.C
+++ b/root/io/filemerger/execTestMultiMerge.C
@@ -79,7 +79,11 @@ int execTestMultiMerge()
    result += testSimpleFile("hsimple.root",25000,1,414415, kIs32bits ? 10 : 8);
    result += testSimpleFile("hsimple9.root",25000,9,432029,3);
    // Increasing tolerance due test fail for fast-math builds and i686
-   result += testSimpleFile("hsimple101.root",25000,101,414590, kIs32bits ? 10 : 8);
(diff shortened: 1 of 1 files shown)


root/io/filemerger/execTestMultiMerge.C
```diff
@@ -70,19 +70,22 @@ int testSimpleFile(const char *filename, Long64_t entries, Int_t compSetting, Lo
 int execTestMultiMerge()
 {
    Int_t result = 0;
-   int hsimpleFTolerance = 6; // 5 for non fst builds
+   int hsimpleFTolerance = 10; // 5 for non fst builds
```
pcanal: Do we have a way to detect if were are in a 'fst' build?

oshadura: @pcanal it could defined R__FAST_MATH, I will push new version of fix

oshadura: @pcanal sorry, I mistaken hsimpleFTolerance is for hsimpleF.test case (it is failing only on not-fast aarch64 native compiler build (gcc4.8.5))

oshadura:
@pcanal, I introduced as you ask separation of fast, i686 builds and others.Can you look please?

PR #73, opened 2017-08-30
Explicitly test LZ4 in the existing compression tests.

bbockelm:
Sadly, when I forcibly create corruptions in the files (causing the LZ4 checksum to fail), the unittest crashes instead of exiting cleanly.

Regardless, corruptions cause this test to fail.

This tweak was done for https://github.com/root-project/root/pull/907 -- but can go in independently of that PR.

@pcanal


bbockelm:
@phsft-bot - build!

pcanal:
@phsft-bot build

PR #91, opened 2017-09-26
Changes in tolerance for files in filemerger test

oshadura:
Updating tolerance due changes in default value of fTitleOffset for ROOT v6-08-00


oshadura:
@pcanal I rechecked pedestals! Can you check please if it is ok?

PR #93, opened 2017-09-27
Increasing tolerance for aarch64 fast-math builds (different precision)

oshadura:


oshadura:
@pcanal can you take a look please?

oshadura:
@pcanal thank you for such elegant idea, i checked on aarch64 with fast math, test is green!

oshadura:
@amadio PR with new tolerances!

amadio:
@oshadura Let me merge the fix for the test I broke last night, then we can proceed here. (If the references are not from yesterday after my merges or from today, they should be ok, though).

PR #94, opened 2017-09-28
Updating tolerance for i686 build (changed default value of fTitleOffset)

oshadura:


oshadura:
@pcanal it is tiny fix , after today nightlies..

pcanal:
Could you update the commit message? (It seems that the commits changes the tolerance rather than the pedestal).

oshadura:
@pcanal done!

PR #97, opened 2017-10-05
Cleaning out old version of Googletest, adding SearchInstalledSoftwar…

oshadura:
Cleanup of code(version of gtest was 1.7.0 ~ 3 years old), adding SearchInstalledSoftware.cmake with new version of GTest


oshadura:
Closing PR, I didn't test it as a part of ROOT (-Droottest=ON)

oshadura:
@vgvassilev I checked both standalone and build as a part of ROOT, looks like working! Thanks for tip!

oshadura:
@phsft-bot build!

oshadura:
@phsft-bot build!

oshadura:
@pcanal Can you take a look please? I cleaned a bit old version of gtest and added download of latest gtest version as a part of SearchInstalledSoftwareRoottest.cmake.

It works also well in standalone mode...

root/hist/chain/CMakeLists.txt
```diff
@@ -1 +1,2 @@
+include_directories(${GTEST_INCLUDE_DIR})
```
pcanal: Why does this need to be repeated in each CMakeList.txt?

oshadura: Not in every CMakeLists.txt, but only where is used gtest, before gtest includes were inside source tree and was easily located, now we need to add them to includes, no?

oshadura: Do you want to move it in ROOTTEST_ADD_UNITTEST_DIR macro?

pcanal: Well 'sort of'.  The include_directories(${GTEST_INCLUDE_DIR}) should be done at the global level in a similar place that induce the definition of ROOTTEST_ADD_UNITTEST_DIR  to be available (without any explicit include) in each CMakeList.txt.

oshadura:
@pcanal, failed build on mac1012 is due infrastructure problems..

PR #101, opened 2017-10-22
Update the reference `TTree::Print` outputs.

bbockelm:
When `TIOFeatures` was serialized as part of the `TTree` and `TBranch`, it changed the number of bytes in the files -- requiring an update to the relevant tests.

Should fix the test failures in https://github.com/root-project/root/pull/1217

@pcanal


pcanal:
@phsft-bot build

PR #102, opened 2017-10-31
Make const methods const in io/stdarray/arrayHolder.h

bluehood:


bluehood:
On a second thought we probably cannot break the interface this way right? People can use TTreeReaderArray's iterators to modify its contents and they should be able to continue doing so.

PR #103, opened 2017-11-04
Remove geom tests if geo is off

dpiparo:


amadio:
LGTM, can you please merge, or should I do it?

PR #105, opened 2017-11-09
Test tree/chain friend support in TTreeProcessorMT

etejedor:
This PR tests the new functionality added to TTreeProcessorMT to take into account the information about tree/chain friends.
This also removes a test that is no longer necessary since TTreeProcessorMT uses now internally a chain instead of an array of files, which means TTreeProcessorMT does not manually switch from one file to another depending on the task entry range, relying instead on the chain to do that work.


etejedor:
The reason for the failing test is of course that it didn't run against a TTreeProcessorMT with friend support.

PR #107, opened 2017-11-10
Fix unnamed macros

vgvassilev:


cling/bytecode/arrayhist.C
```diff
@@ -2,16 +2,14 @@
 gROOT->Reset();
 char name[4][50];
 
-for(int i=0; i<4; i++)
-{
```
pcanal: Isn't this common style supported in unnamed macro? [And if not, is the 'error' message clear]?

If it is not supported, I think this will be surprising to the user (and it is likely to feel very arbitrary to them), so we would to really print a very nice clear message .....

vgvassilev: The idea of these changes is to be consistent with the ROOT prompt: this is currently unsupported in the prompt. IIUC, the unnamed macros should operate as if the user is typing at the prompt.

PR #113, opened 2017-11-28
Provide testing of subdirectories in TXMLFile

linev:
Will work only after apply of https://github.com/root-project/root/pull/1386


linev:
@phsft-bot build

PR #116, opened 2017-12-05
Do not rely on divisions in dataframe reports test

amadio:
There are two kinds of instructions that are used for divisions, divss
and rcpss, which have different accuracy and may be picked up depending
on compiler optimization options. Using multiplications ensures that
division innacuracies will not affect the pass rates of the filters in
this test.


amadio:
Network failure (lost connection).

amadio:
The reference might still need to be updated, but I cannot know without letting this run in other builds. If it does need updating, I will update it.

amadio:
@bluehood You're welcome!

PR #118, opened 2017-12-15
Testing reading objects data from JSON

linev:
Main idea of tests now:
- create JSON and dump it once into ref file
- try to read object from given JSON string
- produce JSON second time from the read object
- compare first and second JSON strings

Test classes are modified - now more non-empty values are present

Will work after merging following PR

https://github.com/root-project/root/pull/1462


pcanal:
@phsft-bot build

PR #120, opened 2017-12-18
Test fixes for ROOT PR1468

dpiparo:


etejedor:
@dpiparo does this still make sense?

PR #126, opened 2018-01-19
Transform old Thread tutorials in roottests

dpiparo:
see https://github.com/root-project/root/pull/1531


etejedor:
Hi @dpiparo we want this in, right?

dpiparo:
@etejedor, no, you can close it without merging.

PR #129, opened 2018-01-26
Update references for root/io/filemerger test

amadio:
The references shifted due to a change introduced by commit
4d26ab79cae22e598834dd2b95d4f5b266be8589 in root.git.


amadio:
Original commits were reverted in ROOT, closing this as it's not necessary anymore.

PR #131, opened 2018-02-09
Wrap ctest command by watcher that sends SIGUSR2 before the timeout.

Axel-Naumann:


cmake/modules/RootCTestMacros.cmake
```diff
@@ -309,6 +309,8 @@ function(ROOTTEST_ADD_TEST testname)
     endif()
   endif()
 
+  set(command "${CMAKE_PROJECT_SOURCE_DIR}/scripts/watch.py ${timeout} -- ${command}")
```
pcanal: You may have meant:  ${CMAKE_SOURCE_DIR}  (without the _PROJECT_) .... at least in some case according to jenkins ${CMAKE_PROJECT_SOURCE_DIR} ends up empty.

Axel-Naumann:
Superseded by #168 !

PR #133, opened 2018-02-21
Correct watch script and impose a 270 seconds timeout

dpiparo:


dpiparo:
@cernphsft build please

PR #135, opened 2018-03-01
Preparing tests for LZ4-based references

oshadura:
Tests: tree-cloning was moved to CMake and was added LZ4 references; tree-cache, io-treeForeign,io-filemerger and io-compression was moved to LZ4 references


root/io/compression/MainEvent.cxx
```diff
@@ -305,10 +305,14 @@ int main(int argc, char **argv)
       int expectedSize = -1;
       if (comp == 0) expectedSize = 5534559;
       else if (comp == 101) expectedSize = 1254808;
-      else if (comp == 201) expectedSize = 1111230;
+      else if (comp == 208) expectedSize = 1111230;
       else if (comp == 301) expectedSize = 1265018;
-      else if (comp == 405) expectedSize = 1284420;
-      else if (comp == 5) expectedSize = 1208369;
+      else if (comp == 404) expectedSize = 1284420;
+#ifdef R__HAS_LZ4
```
pcanal: Does R__HAS_LZ4 means R__HAS_DEFAULT_LZ4?

root/io/compression/MainEvent.cxx
```diff
@@ -305,10 +305,14 @@ int main(int argc, char **argv)
       int expectedSize = -1;
       if (comp == 0) expectedSize = 5534559;
       else if (comp == 101) expectedSize = 1254808;
-      else if (comp == 201) expectedSize = 1111230;
+      else if (comp == 208) expectedSize = 1111230;
```
pcanal: Surprisingly the compression level change but not the final size, is that correct?

oshadura: Thank you, I updated references while running them zero tolerances.

root/io/compression/MainEvent.cxx
oshadura: Updated in both root.git and roottest.git.

PR #139, opened 2018-03-10
Fixing parallel builds for roottest

oshadura:
Add missing dependency for hsimple.
In highly parallel builds we end up executing hsimple before ROOT is fully built. 
This patch fixes a regression seen in the master.


oshadura:
Failure is not connected to PR changes (roottest-root-multicore-processExecutorH1Test).

PR #141, opened 2018-03-12
Tuning of filemerger test  and exectrim references after migration to default lz4

oshadura:


oshadura:
@pcanal can you please take a look? It solved problems with 2 tests in case of lz4 (i636 and tolerance mismatch for hsimple9.root case)

PR #143, opened 2018-03-14
Updating references for tests according final transition of ROOT to L…

oshadura:
…Z4 compression algorithm


pcanal:
@phsft-bot build

oshadura:
@pcanal build was successful, but I saw in incrementals one more test failing that I didn't notice before, so I just updated files and references for it too.

pcanal:
The two error are network timeouts.  Merging.

PR #144, opened 2018-03-14
Moving io-evolution-versions test to LZ4

oshadura:


oshadura:
@pcanal it is a last failing test for LZ4, it will be nice to have it before nightlies

pcanal:
thanks.

PR #146, opened 2018-03-15
json: replace TRef class by TNamed class in testing

linev:
TRef has custom streamer and not correctly supported by JSON I/O

### DIFF
--- a/root/io/json/STL.ref
+++ b/root/io/json/STL.ref
@@ -1916,7 +1916,7 @@ TJsonEx12 store/read/store MATCHED len1:40441 len2:40441
 }
 
  ====== More exotic STL containers as set or map with TRef inside TJsonEx13 ===== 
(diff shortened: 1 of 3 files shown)


pcanal:
Note that this does not solve the problem seen in roottest-root-io-json as the problem is actually that sometimes some elements of the vector are mis-detected as being allocated by TObject::operator new (because the memory allocated by the vector happens to be filled with the magic pattern TObject::operator new).

<https://epsft-jenkins.cern.ch/job/root-nightly-master/BUILDTYPE=Release,COMPILER=native,LABEL=mac1012/1657/testReport/projectroot.roottest.root.io/json/roottest_root_io_json_make/>

pcanal:
However since json-io does not yet fully support TRef, I am merging this nonetheless.

PR #147, opened 2018-03-15
Fixing ninja warning and improving CMake file (hsimple.c is actually …

oshadura:
…returning pointer and not 0)

### DIFF
--- a/root/tree/cloning/CMakeLists.txt
+++ b/root/tree/cloning/CMakeLists.txt
@@ -48,9 +48,11 @@ if(TARGET hsimple)
 endif()
 
 add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/hsimple.root
-                  COMMAND ${ROOT_root_CMD} -q -l -b ${ROOT_DIR}/tutorials/hsimple.C -e return > log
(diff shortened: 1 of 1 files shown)


root/tree/cloning/CMakeLists.txt
```diff
@@ -40,23 +40,24 @@ else()
...
 
 if(TARGET hsimple)
       set(HSimpleDependencies "hsimple")
       add_dependencies(hsimple-file ${HSimpleDependencies})
 endif()
 
-add_custom_command(OUTPUT ${ROOT_DIR}/tutorials/hsimple.root
-                  COMMAND ${ROOT_root_CMD} -q -l -b ${ROOT_DIR}/tutorials/hsimple.C > hsimple.out
+add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/hsimple.root
+                  COMMAND ${ROOT_root_CMD} -q -l -b ${ROOT_DIR}/tutorials/hsimple.C -e return > log
```
pcanal: what about adding to the 'return' some verification that the file was properly created?

oshadura: @pcanal you mean to check if the file exist?

PR #151, opened 2018-03-21
Changing name of logfile to less generic

oshadura:


pcanal:
Failure are just typical timeout.

PR #154, opened 2018-04-03
[TDF] Adapt snapshot test to list of branches with dotted names

etejedor:
Change the expected output of `roottest-root-dataframe-test_snapshot` in order to adapt to the changes proposed in [this PR](https://github.com/root-project/root/pull/1764), related to the support of branches with dotted names.

### DIFF
--- a/root/dataframe/test_snapshot.ref
+++ b/root/dataframe/test_snapshot.ref
@@ -10,10 +10,14 @@ Jitted branch: b1_square
 Jitted branch: b2_vector
 Jitted branch: a
 Jitted means:49 49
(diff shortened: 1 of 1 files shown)


bluehood:
Hi @etejedor , I don't think this is correct: `b1.b1` and `b1` are the exact same quantity, we should not snapshot it twice with two different names.

If this happens, it is an issue of `Snapshot`, not of the test.

etejedor:
Hi @bluehood, absolutely. As discussed in private, this test can be left as it is, but we need another PR that modifies `Snapshot` to not rely on the (possibly dotted) list of branches that we generate now, but on the list of top level branches (`TTree::GetListOfBranches`)

PR #160, opened 2018-04-20
[TDF] Adapt tests to the new report action

dpiparo:


bluehood:
@dpiparo I think you have an extra commit in the PR -- weird that jenkins does not say it conflicts with master

PR #166, opened 2018-04-25
[TDF] Add test for progressive reading of CSV files

etejedor:
This test checks the memory consumed when creating a TDF from a CSV data source in these two cases:
1. When the CSV data source reads the whole content of the CSV file into memory at once.
2. When the CSV data source is constructed with a chunk size parameter, so that the reading of the rows is progressive.

In the second case, the memory consumed should be much less than in the first, given the chunk size with which the data source is built.


etejedor:
This should not be merged before [this PR](https://github.com/root-project/root/pull/1831) goes in.

etejedor:
@bluehood done!

etejedor:
Testing now that the corresponding functionality has been merged into ROOT master.

etejedor:
@phsft-bot build

PR #170, opened 2018-05-04
Fix tests for modules; Add include header guards for payloads

yamaguchi1024:


Axel-Naumann:
I'm curious to understand why those are needed!

(And include guards must not start with `__`.)

PR #172, opened 2018-05-07
[cxxmodules] Fix a test for modules

yamaguchi1024:
Follow up patch for 25b8f34a1f98b9af1199b5b27bcbff9870c4d9b6, as it was
excluding ROOT_MODULES which was cxxmodules not runtime_cxxmodules.


vgvassilev:
Failures are unrelated.

PR #174, opened 2018-05-08
Fix runtime modules tests

vgvassilev:


vgvassilev:
Looks like our infrastructure fails inform GH. Otherwise looks green.

PR #190, opened 2018-05-29
[cxxmodules] Do not just load all libraries

yamaguchi1024:
We have a mechanism in both modules and pch to load libraries on-demand,
   so no need to "preload all libraries"


Axel-Naumann:
This test intentionally loads all libraries. This change should not go in.

PR #195, opened 2018-06-02
Adjust file size tolerances to reflect improvements in LZ4 1.8.2

jblomer:
Fixes a test case failure when ROOT is built against liblz4 1.8.2.  The ROOT files where the file size is different have been compared with the ones produced by a built-in LZ4 ROOT using [a macro to convert the contents to JSON](https://github.com/jblomer/hjson/blob/roottest-fix-lz4-1.8/hjson.C).

PR #198, opened 2018-06-07
Adjust file size tolerances to reflect improvements in LZ4 1.8.2

jblomer:
Resolves conflicts from #195


pcanal:
This PR affect only roottest.root.io.filermerger, the error reported are unrelated (mix of pre-existing, random and infrastructure).

PR #200, opened 2018-06-08
Fix tpprocessimt v614

bluehood:


bluehood:
the test `tp_process_imt` failed [on multiple platforms](in the latest nightlies).
merged this PR as it has fixed the problem in master.

PR #202, opened 2018-06-11
[PyROOT] Fix operator test with new str behaviour

stwunsch:
Fixes PyROOT test of basic operators, which compares `str(...)` output to expectation. The behaviour changes with the new pretty-printing feature [in this PR to ROOT](https://github.com/root-project/root/pull/2097).

The changes are backward compatible because `repr(...)` is similar to `str(...)` in the old behaviour.


stwunsch:
@phsft-bot build

Axel-Naumann:
> The changes are backward compatible because repr(...) is similar to str(...) in the old behaviour.

Well, "similar behavior" is not how we traditionally define backward compatibility ;-) But I don't think this matters - a usability enhancement is dfinitely worth it :-)

PR #222, opened 2018-09-30
[cxxmodules] Stabilize the test.

vgvassilev:
Copy and move ctors are lazily synthesized by clang on demand. Runtime cxxmodules
introduce yet another step of lazyness which causes this test to fluctuate.

We have seen fluctuations with this test when we upgrade llvm.

Patch by Axel Nauman and me!


vgvassilev:
@phsft-bot build just on slc6/gcc62 with flags -Druntime_cxxmodules=On

vgvassilev:
This is solved by root-project/root#5036 and the test should be reenabled by #450.

PR #224, opened 2018-10-02
Use system version of pytest and pytest_cov

amadio:
The bundled version is too old, conflicts with system installed pytest, and leads to spurious test failures.


amadio:
@phsft-bot build

PR #226, opened 2018-10-03
[cmake] Make roottest independent on ROOT_SHOW_OPTIONS.

vgvassilev:
This allows roottest to be checked out outside ROOT's source tree. Now we
rely on the list of options/features provided by the ROOTConfig.cmake.

This patch should fix ROOT-9405.

### DIFF
--- a/cmake/modules/RootCTestMacros.cmake
+++ b/cmake/modules/RootCTestMacros.cmake
@@ -97,7 +97,6 @@ macro(ROOTTEST_SETUP_EXECTEST)
 
 endmacro(ROOTTEST_SETUP_EXECTEST)
 
-
 #-------------------------------------------------------------------------------
 #
 # function ROOTTEST_ADD_TEST(testname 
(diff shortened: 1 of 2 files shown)


bluehood:
Hi @vgvassilev @amadio, @JavierCVilla and I were trying to build roottest against an existing installation of ROOT and found out we also need this patch. It would be great to have it in master :smile:

bluehood:
By the way, if I understand correctly, it's not true that "This allows roottest to be checked out outside ROOT's source tree" -- that's how I use roottest all the time, _but_ roottest must then be built together with ROOT.

bluehood:
@vgvassilev @amadio ping

cmake/modules/RootCTestMacros.cmake
```diff
@@ -131,24 +129,28 @@ function(ROOTTEST_ADD_TEST testname)
   endif()
 
   if (ARG_ENABLE_IF OR ARG_DISABLE_IF)
-    ROOT_SHOW_OPTIONS(ROOT_ENABLED_FEATURES)
     # Turn the output into a cmake list which is easier to work with.
-    STRING(REPLACE " " ";" ROOT_ENABLED_FEATURES "${ROOT_ENABLED_FEATURES}")
+    set(ROOT_ENABLED_FEATURES ${_root_enabled_options})
+    set(ROOT_ALL_FEATURES ${_root_all_options})
```
amadio: This variable is not set anywhere that I can see, I checked root.git,  roottest.git, and installed files from ROOT.

vgvassilev: This variable is part of PR root-project/root#2122

PR #227, opened 2018-10-03
Fix unnamed macros

vgvassilev:


vgvassilev:

cling/bytecode/arrayhist.C
```diff
@@ -2,16 +2,14 @@
 gROOT->Reset();
 char name[4][50];
 
-for(int i=0; i<4; i++)
```
pcanal: Can you open a JIRA ticket to remind us that this syntax (and the one without any curly) is not yet supported and might be supported with additional clang feature?

root/collection/execTClonesArrayAbsorb.C
```diff
@@ -1,5 +1,5 @@
-{
-#ifdef ClingWorkAroundMissingImplicitAuto
+void execTClonesArrayAbsorb() {
+#ifndef ClingWorkAroundMissingImplicitAuto
```
pcanal: The reversion seems wrong.  The declaration needs to be present when the user request the workaround 'missing implicit auto keyword'.

pcanal: on the other hand there is no 'implicit auto' unless we have an unnamed macro.

pcanal: so the question really is, "is implicit now completely working?" (i.e. should we remove the #ifdef).

root/collection/execTClonesArrayAbsorb.C
```diff
@@ -1,9 +1,6 @@
 void execTClonesArrayAbsorb() {
-#ifndef ClingWorkAroundMissingImplicitAuto
-   TClonesArray *tc, *tc2;
-#endif
-tc = new TClonesArray("TNamed",5); for(int i=0; i<20; ++i) tc->ConstructedAt(i);
-tc2 = new TClonesArray("TNamed",5); for(int i=0; i<20; ++i) tc2->ConstructedAt(i);
+TClonesArray *tc = new TClonesArray("TNamed",5); for(int i=0; i<20; ++i) tc->ConstructedAt(i);
```
pcanal: > The test is supposed to test leaks in TClonesArray, not implicit auto. 

Do we have a test or JIRA ticket that capture **this** use case?  (I.e. when we work on fixing implicit auto, are we reducing the test coverage by jettison this use case?)

root/treeformula/stl/execReferenceReturnType.C
```diff
@@ -7,6 +7,9 @@ v.emplace_back("a","b");
 t.Fill();
 t.Scan("v@.at(0).GetName()");
 auto res = t.Scan("v@.at(0).fName");
-if (res == 1) return 0;
-else return 1;
+if (res == 1) {
```
pcanal: > Cling's continuation does not support this syntax. 

Do we have a JIRA ticket following this mis-use and/or documentation that it will not be supported?

vgvassilev:
I have closed the root.git PR concerning this PR. I am closing this now and reopen when I have bandwidth.

PR #228, opened 2018-10-03
Test behavior in TClusterIterator when there are clusters but fAutoFl…

pcanal:
…ush is still negative.


pcanal:
The new test is failing as expected (i.e. demonstrating the deficiency).

PR #232, opened 2018-10-08
Add test that weak symbols do not trigger autoloading.

Axel-Naumann:
Also, test that once the library is loaded, weak symbols are
resolved from the library instead of recompiling them in the JIT.


vgvassilev:
@phsft-bot build!

PR #235, opened 2018-10-11
Update LZ4 test to reflect change in compression ratio.

bbockelm:
Corresponds to https://github.com/root-project/root/pull/2775

Looks like setting `fLenType` in the `TLeafElement` resulted in a very small compression ratio change in one of the unittests.  This is the corresponding change.


pcanal:
@bbockelm it is not clear why the test is not failing here ...

PR #237, opened 2018-10-18
Temporary disabling roottest_root_treeformula_sync_make test for rt c…

oshadura:
…xxmodules

Error is visible only on Jenkins and non deterministic:
--- Using ROOT from /mnt/build/jenkins/workspace/root-pullrequests-build/build
./copiedEvent: symbol lookup error: ./libEvent.so: undefined symbol: _ZN5TROOT14RegisterModuleEPKcPS1_S2_S1_S1_PFvvERKSt6vectorISt4pairISsiESaIS7_EES2_b

### DIFF
--- a/root/treeformula/sync/CMakeLists.txt
+++ b/root/treeformula/sync/CMakeLists.txt
@@ -4,4 +4,6 @@
 # define a CTest test that calls 'make' in ${CMAKE_CURRENT_SOURCE_DIR}
 #
 #-------------------------------------------------------------------------------
-ROOTTEST_ADD_OLDTEST()
+if(NOT ROOT_runtime_cxxmodules_FOUND)
+  ROOTTEST_ADD_OLDTEST()
+endif()


pcanal:
The error mentioned happens also without the runtime module.  See for example https://github.com/root-project/root/pull/2807 and https://epsft-jenkins.cern.ch/job/root-pullrequests-build/44825/testReport/projectroot.roottest.root.treeformula/sync/roottest_root_treeformula_sync_make/

This is likely a missing dependency of sort.  @amadio can you take a look?

pcanal:
@oshadura @amadio This is a bit confusing but fair enough.  thanks.

oshadura:
@pcanal I know and I am sorry for this, I will try to add a message when the build failed, with which flags it was running.

amadio:
The pull requests don't have a labeling mechanism as the nightlies, so we have to live with the fact that some settings are not explicit (i.e. C++17, modules, etc).

oshadura:
Vassil gave me a nice idea just to write in a body of message, flags which was used to build, I will try to do this then.

PR #240, opened 2018-10-24
[WIP] Updating test references in case when is used builtin_zlib

oshadura:


oshadura:
@pcanal I am still not ready, I will ping you later..

PR #245, opened 2018-10-30
Reduce size for 32bit reference for LZ4-9 (lz4 1.8.3)

oshadura:


oshadura:
@phsft-bot build with flags -Dctest_test_exclude_none=on

oshadura:
@pcanal somehow reference values for clang39 is 2 bytes bigger, which could be similar to https://sft.its.cern.ch/jira/browse/ROOT-9191.
32b build is fixed.

pcanal:
Did you understand ROOT-9191?  Do we know why it is only sporadic?

oshadura:
@pcanal, let me merge this PR since it is fixing 32 bit problem with references for lz4 1.8.3, and I want to try if it is only problem in clang39 or it is visible in other version of clang. I will open new pr with fix.

@Axel-Naumann  what is our policy with clang 3.9? we need to use it because it is default clang compiler available centos7 repos?

Axel-Naumann:
We can kill clang 3.9.

PR #247, opened 2018-10-31
[ZLIB] Back to zlib as default algorithm

oshadura:
It is a synchronization of revert commits with branch 6.14 patches.
Next PR will generalize a switch between compression algorithms.


oshadura:
I am testing both branches here: https://github.com/root-project/root/pull/2889

PR #250, opened 2018-11-05
[TTreeReader][ROOT-9757] Check number of branches read

JavierCVilla:
Add test for [#2921](https://github.com/root-project/root/pull/2921) in root project.


dpiparo:
This is for me ready to go. Thanks @JavierCVilla !

PR #254, opened 2018-11-22
[DF] Fix library dependencies in CMakeLists

bluehood:


root/dataframe/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ if(ROOTTEST_OS_ID MATCHES Ubuntu)
   set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--no-as-needed")
 endif()
 
-set(DFLIBRARIES Core ROOTDataFrame Hist Tree RIO MathCore)
+set(DFLIBRARIES Core Hist RIO MathCore Tree TreePlayer ROOTVecOps ROOTDataFrame)
```
amadio: Hi Enrico, don't these extra libraries come automatically from linking to ROOTDataFrame? If not, we should try to fix what is exported so that users don't have to specify dependencies by hand.

bluehood: Apparently not: try having a ROOT installation, then compile rootttest and run for instance `ctest -R roottest-root-dataframe-test_columnoverride --output-on-failure`

amadio: Ok, than I think it's better to fix this in ROOT proper so that the extra libraries are not needed here. Note that the dependencies of RDataFrame may change depending on configuration, so we should export the right things. For example, there may be an optional dependency on SQLite or Arrow for their data sources, we may make the dependency on TreePlayer go away in the future, etc...

bluehood: I don't see anything wrong in `$ROOTSYS/tree/dataframe/CMakeLists.txt` though....help?

amadio: You are right, the bug is in `ROOT_LINKER_LIBRARY()` when used from an external project. See [ROOT-9827](https://sft.its.cern.ch/jira/browse/ROOT-9827) for details. Let's fix this in ROOT and close this PR. Thanks for finding the bug!

PR #255, opened 2018-11-23
Fix rtcxxmod: Exp used as different symbol in oce/Standard_Real.hxx.

Axel-Naumann:


Axel-Naumann:
`Exp` is already used as function in OpenCascade, and clashes with the use as namespace in this test (when both are loaded, as is the case for runtime modules). Because this test uses old files, it's not enough to simply update the code. @pcanal please let me know what you prefer to do to fix this.

vgvassilev:
Ping.

pcanal:
See https://github.com/root-project/roottest/pull/337 (which also change the legacy file).

PR #257, opened 2018-11-27
Fix call to Ast Call in pytest for Python3

etejedor:
This PR ports a fix to ROOT's pytest to get rid of these error:

http://cdash.cern.ch/testDetails.php?test=57901667&build=598659

The fix has been picked from a pytest PR:

https://github.com/pytest-dev/pytest/pull/801

The objective is to properly handle the calls to Ast Call depending on the Python version, since its signature changed in Python3.5.


Axel-Naumann:
@phsft-bot build just on ROOT-fedora27/noimt

etejedor:
@phsft-bot build just on ROOT-fedora27/noimt

PR #259, opened 2018-11-29
Fixes to Rules.mk

bluehood:


bluehood:
@Axel-Naumann I'm appending rather than prepending to CXXFLAGS so that ROOT's flags take priority -- is that ok?

scripts/Rules.mk
```diff
@@ -322,9 +322,7 @@ ifeq ($(ROOTBITS),)
    export ROOTBITS := $(shell root.exe -b -q -n $(ROOTTEST_HOME)/scripts/Bits.C | grep Bits_in_long | awk '{print $$2;}' )
 endif
 
-ifeq ($(CXXFLAGS),)
-   export CXXFLAGS := $(shell root-config --cflags)
-endif
+export CXXFLAGS := "${CXXFLAGS} $(shell root-config --cflags)"
```
pcanal: One challenge is that the makefile execution are recursive while drilling down the sub-directories.  This is (likely to be) adding the root flags for each sub-directory.

bluehood: ah, didn't know that :sweat_smile: 

the problem is that if `CXXFLAGS` is already set in the user environment, as things are we never add `root-config --cflags` to it. Any idea what would be a good fix?

Axel-Naumann:
```
ifeq($(WE_HAVE_ADDED_ROOT_CFLAGS_TO_CXXFLAGS,))
  export CXXFLAGS := "${CXXFLAGS} $(shell root-config --cflags)"
  WE_HAVE_ADDED_ROOT_CFLAGS_TO_CXXFLAGS:=1
endif
```
?

bluehood: can do!

pcanal: Indeed, we already do this a few places we the prefix ROOTTEST_, eg. ROOTTEST_CHECKED_VERSION

bluehood:
@pcanal this seems to fix the original issue (and not cause havoc otherwise), I will merge this afternoon if that's ok

PR #262, opened 2018-11-30
[DF] Use ROOT::ROOTDataFrame instead of ROOTDataFrame as a dependency

bluehood:
The former correctly brings in all of RDataFrame's library dependencies
also in the case of roottest being built against an already installed
version of ROOT.


bluehood:
ROOT doesn't know `ROOT::ROOTDataFrame`, PR marked as "do not merge" for now

bluehood:
@amadio I don't remember, is there a jira ticket I can track for the resolution of this problem?

PR #265, opened 2018-12-07
Replace use of rmkdepend with standard makedepend

amadio:
Just trying this out to see if the standard tool would work just fine for the few tests that still use it.


amadio:
@phsft-bot build

PR #273, opened 2019-01-15
[Exp PyROOT] Flag tests that now fail in PyROOT experimental

etejedor:
A number of test failures have to be fixed in the experimental PyROOT builds:

https://epsft-jenkins.cern.ch/job/root-exp-pyroot/9/

This PR temporarily flags those tests as "will fail" for the experimental PyROOT builds, and they will be restored progressively as they are fixed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,6 +117,11 @@ if(ROOT_python_FOUND)
   find_package(PythonLibs)
 endif()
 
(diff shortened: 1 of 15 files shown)


amadio:
You need to set the variable here independently of ROOT, or it will only work when roottest is built as part of ROOT. Other than that, seems ok.

PR #281, opened 2019-01-31
[DF] Add TreePlayer as an explicit dependency for tests

bluehood:
Otherwise, when building roottest against a pre-existing ROOT
installation, libTreePlayer is not linked.
This explicit dependency on libTreePlayer can be removed once the
ROOTDataFrame target transitively brings in its dependencies also
when building roottest against a pre-existing ROOT installation.

@amadio how do we test that this resolves the failures you were seeing?


amadio:
I guess right now we'd have to try locally, since there's only my build testing roottest outside of ROOT. Once this is merged, we can trigger my build again and check that the tests are fixed.

PR #294, opened 2019-02-14
Adjust tolerance for 32 build builds

oshadura:


oshadura:
Checked on build node, reference adjustment helped.

PR #295, opened 2019-02-18
[cxxmodules] Disable test which assumes existence of rdict files.

vgvassilev:


Axel-Naumann:
@vgvassilev can and should this be merged into master before the v6.18 branch?

vgvassilev:
Nope.

pcanal:
Even-though this test *does* check the content of the rdict files, it is *not* is primary goal (i.e. we should probably find a way to disable the rdict check when they are not available but keep the rest).

Axel-Naumann:
@vgvassilev can this be closed?

PR #296, opened 2019-02-28
Updating references for 32b

oshadura:


oshadura:
Commit is fixing 32 bit failures. projectroot.test.test_TFormulaTest test failure is not connected with introduced changes.

PR #297, opened 2019-03-11
CMake GenerateDict dependency fixes / roottest

Axel-Naumann:
Missing roottest PR for https://github.com/root-project/root/pull/3404


Axel-Naumann:
Was already tested as part of https://github.com/root-project/root/pull/3404

PR #299, opened 2019-03-15
[Exp PyROOT] Make tests compliant with new Cppyy

etejedor:
This PR modifies `roottest-python-basic-basic` and `roottest-python-function-function` to adapt them to the new experimental PyROOT and its new Cppyy. Every commit explains the reason for the corresponding changes.

It is expected that, after these modifications, both tests will still fail. In the case of  `roottest-python-basic-basic`, there is still this issue:
https://sft.its.cern.ch/jira/browse/ROOT-10041

As for `roottest-python-function-function`, there is this remaining issue:
https://sft.its.cern.ch/jira/browse/ROOT-10029

Both issues have already been fixed in the Cppyy repos and will be incorporated in the next Cppyy update for PyROOT experimental.

### DIFF
--- a/python/basic/PyROOT_basictests.py
+++ b/python/basic/PyROOT_basictests.py
@@ -127,7 +127,12 @@ class Basic3PythonLanguageTestCase( MyTestCase ):
    def test3ThreadingSupport( self ):
       """Test whether the GIL can be properly released"""
 
-      gROOT.GetVersion._threaded = 1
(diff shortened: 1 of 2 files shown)


etejedor:
@phsft-bot build with flags -Dpyroot_experimental=On

Tests still pass in current PyROOT, now let's try experimental.

PR #301, opened 2019-03-19
Define CMAKE_MODULE_PATH for roottest.git too

oshadura:
Fix problem visible on macosx1014:
CMake Error at /build/jenkins/workspace/root-pullrequests-build/roottest/CMakeLists.txt:142 (ROOTTEST_ADD_TESTDIRS):
17:47:23   Unknown CMake command "ROOTTEST_ADD_TESTDIRS".

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,6 +67,8 @@ else()# testing using an installation
   enable_testing()
 endif()
 
+set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules)
+
 # Synchronizing default compression algorithm used for ROOT.
 # We need to have it for CMake settings for switching tests references.
 set(compression_default "zlib" CACHE STRING "" FORCE)


Axel-Naumann:
@oshadura can and should this be merged into master before the v6.18 branch? How does this relate to https://github.com/root-project/roottest/pull/305# ?

oshadura:
It is not related to #305, but to https://github.com/root-project/root/pull/3409

amadio:
I guess this is not needed anymore. You probably hit the same problem as me in an earlier version of the PR (i.e. having a `RootMacros.cmake` in ROOT broke roottest, because it started finding the ROOT one and not the roottest one). This sets `CMAKE_MODULE_PATH` to contain only the `roottest/cmake/modules` directory. What we want is probably to just prepend to `CMAKE_MODULE_PATH` to make sure that files in roottest are found first. I also created PR #338 to get rid of `RootMacros.cmake` in roottest, so that we can rename `RootNewMacros.cmake` to `RootMacros.cmake` in the future. It doesn't make sense to have it called `RootNewMacros.cmake` given that there is no `RootMacros.cmake` or `RootOldMacros.cmake`, etc.

oshadura:
> I guess this is not needed anymore. You probably hit the same problem as me in an earlier version of the PR (i.e. having a `RootMacros.cmake` in ROOT broke roottest, because it started finding the ROOT one and not the roottest one). This sets `CMAKE_MODULE_PATH` to contain only the `roottest/cmake/modules` directory. What we want is probably to just prepend to `CMAKE_MODULE_PATH` to make sure that files in roottest are found first. I also created PR #338 to get rid of `RootMacros.cmake` in roottest, so that we can rename `RootNewMacros.cmake` to `RootMacros.cmake` in the future. It doesn't make sense to have it called `RootNewMacros.cmake` given that there is no `RootMacros.cmake` or `RootOldMacros.cmake`, etc.

Yes, please! RootNewMacros.cmake is the most annoying thing in ROOT :)

oshadura:
I will close it then

PR #304, opened 2019-04-05
Teach roottest to exclude directories with tests

oshadura:


pcanal:
@phsft-bot build

oshadura:
hmmm, timeout for bulk io tests, i will take a look later why they are so slow.

PR #307, opened 2019-04-12
Change test to check when new lines of CSV file are read

etejedor:
Change test to check when new lines of CSV file are read. The previous strategy, which consisted in checking the resident memory of the process before and after running the RDF computation, was not reliable and made the test fail intermittently.


etejedor:
@phsft-bot build

PR #314, opened 2019-04-17
json: test writing and reading std::map as JSON object

linev:
Testing new funcitonlaity to store special kinds of map as JSON object.
Such representation is much more compact than default one.
But works only for limited usecases, therefore cannot be used by
default.

### DIFF
--- a/root/io/json/runMap.C
+++ b/root/io/json/runMap.C
@@ -37,10 +37,10 @@ void runMap()
       data[key] = n*7;
    }
 
-   testJson(data, 0, "DEFAULT: ");
+   testJson(data, TBufferJSON::kNoCompress, "DEFAULT: ");
 
-   testJson(data, 5, "OBJECT: ");
+   testJson(data, TBufferJSON::kMapAsObject, "OBJECT: ");
 
-   testJson(data, 108, "MINIMAL: ");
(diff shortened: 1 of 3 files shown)


root/io/json/runMap.C
```diff
@@ -0,0 +1,46 @@
...
+
+   for (int n=0;n<10;++n) {
+      std::string key = "field";
+      key.append(std::to_string(n));
+      data[key] = n*7;
+   }
+
+   testJson(data, 0, "DEFAULT: ");
+
+   testJson(data, 5, "OBJECT: ");
```
pcanal: My apologies, I missed this details in the original PR.   Could we add an enum type for the user to be able to use semantic rather than magic number ?

linev: These "magic" numbers are since the beginning of TBufferJSON.
I will add them in separate PR

pcanal: thanks.

linev: I modify test to use introduced constants

pcanal: thanks.

linev:
@phsft-bot build

pcanal:
Windows failure is unrelated.

PR #320, opened 2019-05-10
[cxxmodules] Narrow the excluded tests.

vgvassilev:
We fail only when running roottest/root/io/tuple tests.


Axel-Naumann:
@vgvassilev can and should this be merged into master before the v6.18 branch?

vgvassilev:
It can but not necessary.

PR #325, opened 2019-05-17
C++17 fixes

Axel-Naumann:


python/pythonizations/PyROOT_smartptrtest.py
```diff
@@ -34,7 +34,7 @@ def test01_transparency(self):
         mine = cppyy.gbl.mine
 
         assert type(mine) == MyShareable
-        assert type(mine._get_smart_ptr()) == cppyy.gbl.std.auto_ptr(MyShareable)
+        assert type(mine._get_smart_ptr()) == cppyy.gbl.std.shared_ptr(MyShareable)
```
pcanal: this seems to need more change to the script ....

Axel-Naumann: I see the failures, but I don't understand them :-/

pcanal: Did you change the implementation of _get_smart_ptr?

Axel-Naumann: No. And I shouldn't have to (if you look at the machinery that makes this work - in the end, `Cppyy::IsSmartPtr()`. More guesses? ;-)

pcanal: Then I don't understand the semantic of the line ... if nothing else change why would the return of type(mine._get_smart_ptr()) switch from auto_ptr to shared_ptr?

pcanal: humm .. you also change SmartPtr* ... so okay ...

pcanal: and I was off by one line on where the error happen, so never mind ....

Axel-Naumann:
Needs https://github.com/root-project/root/pull/3840

Axel-Naumann:
@phsft-bot build

PR #332, opened 2019-06-04
[ROOT-10099] Adapt to change in TH* class version

oshadura:


oshadura:
Failures are unrelated to changes, backport is fixing wrong references visible in CDash dashboard.

PR #334, opened 2019-06-05
Fix standalone roottest

oshadura:
I removed dependency on ROOTSYS and now it is enough to have root in PATH for standalone case.
It will allow easier to test binaries.


oshadura:
Failures are not related with changes..

oshadura:
@amadio are you fine with these changes?

PR #338, opened 2019-06-13
Merge RootMacros into RoottestMacros, and update required version of CMake

amadio:


amadio:
Looks like the pull request builds are fetching sources from https://root.cern instead of GitHub. That's most likely wrong.

amadio:
@oshadura There are two tests with the same name in `root/tree/fastcloningeventree`, which make configuration fail when runtime_cxxmodules is ON. Could you take a look? `git blame` points to https://github.com/root-project/roottest/commit/326997dd3e580a236b29e2a90c958d8609f2959f.

oshadura:
@amadio I am so sorry, I think we did the same commits and I landed them before: https://github.com/root-project/roottest/pull/338/commits/2f4fa379c47c8439455a2ae9bc5a19e978001ef7 and https://github.com/root-project/roottest/pull/338/commits/69cc64b5f9ccd39ad14308e28f12eadab6490337

amadio:
Don't worry, I already rebased locally.

PR #340, opened 2019-06-14
[Exp PyROOT] Roottest-python-basic-operator enabled

maxgalli:
Change needed to enable the above mentioned test.


python/basic/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ if(ROOT_python_FOUND)
                     MACRO PyROOT_operatortests.py
                     COPY_TO_BUILDDIR Operators.C
                     PRECMD ${ROOT_root_CMD} -b -q -l -e .L\ Operators.C+
-                    ${PYTESTS_WILLFAIL})
+                    ENVIRONMENT EXP_PYROOT=${exp_pyroot})
```
etejedor: Do you need the `EXP_PYROOT` variable when the test runs?

PR #345, opened 2019-06-27
[cxxmodules] Filter out source locations.

vgvassilev:
If the declaration comes from an AST file no source locations are printed.

This patch is in the same direction as we took in root-project/roottest@3a8dfca4

This is part of root-project/root#3850


vgvassilev:
@phsft-bot build just on ROOT-ubuntu16/rtcxxmod with flags -DCTEST_TEST_EXCLUDE_NONE=On

PR #348, opened 2019-07-04
[Exp PyROOT] Two failing tests fixed: python-pickle-read and write

maxgalli:


python/pickle/PyROOT_readtests.py
```diff
@@ -107,11 +107,15 @@ def __doftest( self, d ):
    def test5ReadCustomTypes( self ):
       """Test reading PyROOT custom types"""
 
+      proto = [ctypes.c_long(123), ctypes.c_double(123.123)]
+
       o = pickle.load( self.in1 )
-      self.assertEqual(o, [123,123.123] )
+      for touple in zip(o, proto):
```
etejedor: Instead of `tuple` you can put here `for e1,e2 in zip(o, proto)` so that Python already binds the list elements to two variables and you do not need to index the tuple afterwards.

python/pickle/PyROOT_readtests.py
```diff
@@ -107,11 +107,15 @@ def __doftest( self, d ):
...
+         self.assertEqual(touple[0].value, touple[1].value)
 
       o = cPickle.load( self.in2 )
-      self.assertEqual(o, [123,123.123] )
+      for touple in zip(o, proto):
```
etejedor: Same comment here.

python/pickle/PyROOT_readtests.py
```diff
@@ -5,7 +5,7 @@
 
 """Pickle writing unit tests for PyROOT package."""
 
-import os, sys, unittest
+import os, sys, unittest, ctypes
```
etejedor: General comment: the modifications to the tests need to be executed only when we are in experimental PyROOT, the old PyROOT test code should remain the same. You need to follow a similar strategy as in the previous test you worked on, where you exported an environment variable to the test and, depending on the value of that variable, you had one behavior or another.

python/pickle/PyROOT_writetests.py
```diff
@@ -75,7 +75,7 @@ def test4WriteROOTObjInNamespace( self ):
    def test5WriteCustomTypes( self ):
       """Test writing PyROOT custom types"""
 
-      o = [ ROOT.Long(123), ROOT.Double(123.123) ]
+      o = [ctypes.c_long(123), ctypes.c_double(123.123)]
```
etejedor: For example, `ROOT.Long` and `ROOT.Double` still need to be tested in the old PyROOT!

python/pickle/PyROOT_readtests.py
```diff
@@ -107,12 +107,27 @@ def __doftest( self, d ):
    def test5ReadCustomTypes( self ):
       """Test reading PyROOT custom types"""
 
-      o = pickle.load( self.in1 )
-      self.assertEqual(o, [123,123.123] )
+      self.exp_pyroot = os.environ.get('EXP_PYROOT') == 'True'
```
etejedor: If you only use `exp_pyroot` in this function, you do not need to bind it to self, just use a local variable.

python/pickle/PyROOT_writetests.py
```diff
@@ -75,7 +75,14 @@ def test4WriteROOTObjInNamespace( self ):
    def test5WriteCustomTypes( self ):
       """Test writing PyROOT custom types"""
 
-      o = [ ROOT.Long(123), ROOT.Double(123.123) ]
+      self.exp_pyroot = os.environ.get('EXP_PYROOT') == 'True'
```
etejedor: Same comment as above.

python/pickle/PyROOT_readtests.py
```diff
@@ -107,12 +107,27 @@ def __doftest( self, d ):
...
+
+      p = pickle.load( self.in1 )
+      cp = cPickle.load( self.in2 )
+
+      if self.exp_pyroot:
+         import ctypes
```
etejedor: Can we add here a small comment saying why we test `ctypes` instead of `ROOT.Long` and `ROOT.Double`?

For example, if you see here:
https://github.com/root-project/roottest/commit/822fb6316a93453b2f4b1bab20a3c33475e70ef0#diff-7dd31c15ec3c7cc0abde12371883f7b9R240

You can use basically a similar comment (with reference to the cppyy 101 ticket included).

python/pickle/PyROOT_writetests.py
```diff
@@ -75,7 +75,14 @@ def test4WriteROOTObjInNamespace( self ):
...
+
+      if self.exp_pyroot:
+         import ctypes
```
etejedor: Same comment as above, since it is a different test it would be good to have the clarification of why we test `ctypes` also here.

python/pickle/PyROOT_readtests.py
etejedor: We need to refer to both long and double here:
```
# Cppyy's Long and Double will be deprecated in favour of
# ctypes.c_long and ctypes.c_double, respectively
```

python/pickle/PyROOT_writetests.py
etejedor: Same comment as above here.

PR #350, opened 2019-07-08
Adjust file size for LZ4 (409 for aarch64)

oshadura:


Axel-Naumann:
Oops, that was against master, too, just like #349. I have cherry-picked it manually into v6-18-00-patches, as 0d163bd7340ed8e2c249f02bf46ae98d5f1fda10

PR #375, opened 2019-08-25
[Exp PyROOT] Build ROOT with multiple Python versions

maxgalli:
Changed CMake variables to set the environment correctly according to the
selected Python version.


maxgalli:
New version coming

PR #378, opened 2019-08-27
[json] add testing of nested STL containers

linev:
Add testing for fix in https://github.com/root-project/root/pull/4290


linev:
@phsft-bot build

PR #382, opened 2019-08-28
[cxxmodules] Enable and disable the module.modulemap file.

vgvassilev:
The test has subtest which are testing the automatic loading and the project
generation in root. In the former case the module.modulemap file needs to be
present. In the latter case needs to be gone so that we can trigger the
generation of a new project.

The approach of this patch is not the best, however it is unclear if we can
split the logic into multiple tests without a slowdown or generating the same
input/output multiple times.

### DIFF
--- a/root/meta/MakeProject/create_makeproject_examples.C
+++ b/root/meta/MakeProject/create_makeproject_examples.C
@@ -2,8 +2,14 @@ R__LOAD_LIBRARY(stl_makeproject_test)
 
 #include <stl_makeproject_test.h>
 
+bool file_exists (const std::string &file_name) {
+   return gSystem->AccessPathName(file_name.c_str(), kWritePermission);
+}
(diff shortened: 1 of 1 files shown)


pcanal:
there are indeed 3 related steps.
a) generate the test root file using the 'real' library
b) run MakeProject without the 'real' library.
c) test that the files generated by MakeProject works.

a 'cleaner' alternative would be to the have the library of step (a) generated in a sub-directory (so as not interfere with step b)) ... but since c depends on b which depends on a ... if we move them to different directories this becomes 'hard' ... maybe ... @amadio any ideas?  (if not the propose solution is fine ....)

vgvassilev:
Let's merge it now and see if it fixes the failures. We can refactor it later if we have better ideas.

amadio:
This seems ok, I will let you know if I see any problems with it.

PR #389, opened 2019-09-04
[PyROOT][ROOT-10222] Add test for non-ASCII strings

etejedor:
Test for https://github.com/root-project/root/pull/4345


etejedor:
Test failure unrelated to this PR (and can't be observed in the builds of the correspoding root PR).

PR #391, opened 2019-09-09
[json] add testing for JSON_asbase feature

linev:
Test how object can be stored as its base class
Test for new features provided https://github.com/root-project/root/pull/4361


linev:
Closing PR while feature will be not provided in ROOT in that form

PR #397, opened 2019-09-14
[cxxmodules] Clean up with a PRECMD.

vgvassilev:
POSTCMD is not executed when the test fails. CMake aborts with FATAL_ERROR.


vgvassilev:
The failures are in span.cxx, I do not see how they can be related to this pull request.

vgvassilev:
The failures are in span.cxx, I do not see how they can be related to this pull request.

PR #407, opened 2019-10-09
Temporary disabling test for C++ modules

oshadura:
This test is only broken during incremental builds. For incrementals, in this test we need to check if modulemap file exists during CMake runtime and select proper way to execute test depending on existence of modulemap file - the work is ongoing


oshadura:
Sadly I cant disable a test because it generates some data needed for other tests.......

PR #410, opened 2019-10-16
[cxxmodules] Simplify test and make it closer to the real use case.

vgvassilev:
This follows up the changes in root-project/root#4553


vgvassilev:
@phsft-bot build!

PR #439, opened 2020-01-31
Add simple base64 test

linev:
It is now used in TBufferJSON, THttpServer, webgui, therefore has to be
tested

### DIFF
--- a/root/core/base64.C
+++ b/root/core/base64.C
@@ -1,0 +1,34 @@
+#include "TBase64.h"
+
+#include <iostream>
+using namespace std;
+
+void base64()
+{
+   TString binary1 = "This is test binary";
+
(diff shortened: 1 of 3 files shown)


root/core/base64.C
```diff
@@ -0,0 +1,34 @@
...
+
+   std::cout << "base64: " << coded << std::endl;
+
+   binary2 = TBase64::Decode(coded);
+
+   if (binary1 == binary2)
+      std::cout << "base64: coding2 match" << std::endl;
+   else
+      std::cout << "base64: coding2 mismatch" << std::endl;
+}
```
pcanal: Could you also return non-zero in case of detected errors?

linev: Done

PR #443, opened 2020-02-07
Exclude TLego from v6 types list

linev:
TLego and TPainter3dAlgorithms were removed from public includes


linev:
Required after merging of https://github.com/root-project/root/pull/4952

PR #454, opened 2020-03-06
[skip-ci][CMake][Exp PyROOT] Revert 86b5579 and 83562a0

maxgalli:


etejedor:
This goes together with https://github.com/root-project/root/pull/5083

PR #456, opened 2020-03-11
Add missing TList.h in hist/operator/runcopy.C macro

linev:
Appears when ROOT build with "less_includes" option


linev:
Trivial change. Even if "less_includes" will not be part of ROOT

PR #457, opened 2020-03-11
add potentially missing include in execMissingCheckSum.cxx [skip-ci]

linev:
May appear after optimizing ROOT includes


linev:
Trivial changes - mergins

PR #481, opened 2020-04-01
[CmdLineUtils] Add tests for rootls and rootmv and namecycles

eguiraud:
Tests ROOT-10599 and ROOT-10654. These tests are supposed to be broken in current master and fixed by https://github.com/root-project/root/pull/5272


python/cmdLineUtils/CMakeLists.txt
```diff
@@ -348,6 +354,19 @@ ROOTTEST_ADD_TEST(SimpleRootmv5CheckOutput
 ROOTTEST_ADD_TEST(SimpleRootmv5Clean
                   COMMAND ${TOOLS_PREFIX}/rootrm -r move5.root
                   DEPENDS SimpleRootmv5CheckOutput)
+
+
+ROOTTEST_ADD_TEST(NameCyclesRootmvPrepareInput
+                  COMMAND root -q -l MakeNameCyclesRootmvInput.C)
```
pcanal, line 360: Typically we use
```
root.exe -b -l -q
```
to save a bit of time :)

eguiraud: fixed, thanks!

PR #483, opened 2020-04-02
[CMake] Make executables link against sanitizer config.

hageboeck:
When ROOT is built with address sanitizier enabled, ROOTTEST exectuables
need to link against ROOT's sanitizer config to build correctly.


hageboeck:
@phsft-bot build

PR #488, opened 2020-04-14
Fix references for filemereger tests 

oshadura:


oshadura:
@pcanal I thought it could be fixed via references, but it is something different and depends on compiler/os (?) I am not sure what is happening here...

PR #489, opened 2020-04-15
Explicitly link GTEST directory

reikdas:
As suggested by @hageboeck to enhance ROOT PR https://github.com/root-project/root/pull/5297

@vgvassilev


hageboeck:
This fixes the problem in RooFit, but I wonder if that's future proof. Adding @oshadura for this:
- If somebody links against `gtest`, I expect they are also using the gtest headers.
- In that case, *linking against gtest* should automatically pull in the includes for the headers, shouldn't it?
  - If yes, something like this should be added:
         target_include_directories(gtest INTERFACE ${GTEST_INCLUDE_DIR})
     If I'm not mistaken, that would automatically get things right for everybody who uses gtest.
  - If no, this problem might come back in case somebody forgets to explicitly add the includes to what they are doing.

oshadura:
@phsft-bot build!

hageboeck:
This might be unnecessary if root-project/root#5366 does the right thing.

PR #492, opened 2020-04-17
First steps to make roottest working on Windows

bellenot:
First make sure configure and build work. Individual tests will be fixed afterwards


bellenot:
> Can't we have (as in roottest/Makefile) a `$(DllSuf)` variable?

Sure! Will do

pcanal:
Maybe one of the CMAKE_*_SUFFIX , for example CMAKE_SHARED_LIBRARY_SUFFIX ?

bellenot:
> Maybe one of the CMAKE_*_SUFFIX , for example CMAKE_SHARED_LIBRARY_SUFFIX ?

Yes, it's what I have done. testing now and will commit soon

CMakeLists.txt
```diff
@@ -70,6 +71,7 @@ if(MSVC)
 
 else()
 
+  set(libsuffix .so)
```
pcanal, line 75:
```suggestion
  # We can not use CMAKE_SHARED_LIBRARY_SUFFIX as on macos it uses the standard dylib rather than the .so
  set(libsuffix .so)
```
assuming that this is the correct reason.

bellenot: right, thanks!

PR #495, opened 2020-04-24
Use $(PYTHON) instead of `python`; determine which one is found.

Axel-Naumann:


Axel-Naumann:
@phsft-bot build just on ROOT-centos8/default

PR #500, opened 2020-05-05
Substitute usages of GetImplicitMTPoolSize with GetThreadPoolSize

eguiraud:
The former is deprecated in favor of the latter.


eguiraud:
Why does the CI not see the deprecation? We should find out before merging this PR.

eguiraud:
The test build happens in another test, and we don't search tests for warnings.

PR #501, opened 2020-05-06
[JupyROOT] Replace CMake checks by runtime checks for test dependencies

etejedor:
The JupyROOT tests have several dependencies:
1. Doctests: IPython
2. Python kernel tests: jupyter (for nbconvert), which also pulls IPython
3. C++ kernel tests: jupyter and metakernel

This PR moves the check for these test dependencies at runtime instead of at build time.


stwunsch:
I've installed jupyter on the fedora30 node, however, the centos7 node is not managed by us. @oshadura Could you have a look?

stwunsch:
@phsft-bot build

stwunsch:
@phsft-bot build

etejedor:
@phsft-bot build

etejedor:
The Mac1015 failures are not related to this PR (observed in others like http://cdash.cern.ch/viewTest.php?onlyfailed&buildid=872695). 

To be checked with @Axel-Naumann why we get again the `sh: ipython3: command not found` error on that node.

PR #502, opened 2020-05-08
Testing rebase error on Github PR

bellenot:
If this works, then I don't know what the problem could be...


vgvassilev:
Is that still active?

bellenot:
> Is that still active?

yes

PR #302, opened 2019-03-20
[multicore] Test move-only types support in TExecutor::Map

anerokhi:

PR #504, opened 2020-05-11
[multicore] Test move-only types support in TExecutor::Map

Axel-Naumann:
Rebased https://github.com/root-project/roottest/pull/302 - does that fix the errors?


guitargeek:
Apparently it still doesn't work. Let's follow up in the original PR:
  * https://github.com/root-project/roottest/pull/302

PR #507, opened 2020-05-11
[JupyROOT] Add envvar to veto tests

stwunsch:
The envvar is set in `root-build.cmake` and added in the PR here: https://github.com/root-project/rootspi/pull/79


stwunsch:
Do we also plan to run the roottest python tests for py2 and py3? Then, we would have to push the check to runtime.

PR #511, opened 2020-05-12
[JupyROOT] Run tests in build directory

etejedor:
Prevent the creation of .ipynb, .C, .pcm, .d and .so temporary
files in the source directory. Such remnants can cause
failures in the JupyROOT tests, e.g:

https://github.com/root-project/root/pull/5577#issuecomment-625671277

### DIFF
--- a/python/JupyROOT/CMakeLists.txt
+++ b/python/JupyROOT/CMakeLists.txt
@@ -11,11 +11,11 @@ set(DOCTEST_LAUNCHER ${CMAKE_CURRENT_SOURCE_DIR}/doctest_launcher.py)
 # List of notebook files
 # TODO: To be extended with the list of downloaded notebooks used in the
 # documentation and as tutorials.
-set(NOTEBOOKS ${CMAKE_CURRENT_SOURCE_DIR}/importROOT.ipynb
-              ${CMAKE_CURRENT_SOURCE_DIR}/simpleCppMagic.ipynb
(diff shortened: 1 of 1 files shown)


linev:

linev:
@etejedor, I guess same changes should be back-ported to 6.18, 6.20 and may be 6.16 branches
Otherwise nightly tests will continue pollute source dirs
Afterwards one need to go over all build nodes and cleanup source dirs.
@Axel-Naumann, are there tools for that?

Axel-Naumann:
>  Afterwards one need to go over all build nodes and cleanup source dirs.

I suggest we do that on a case-by-case basis, once a node is actually full?

linev:
> I suggest we do that on a case-by-case basis, once a node is actually full?

Nodes are not really full, but just huge number of created files in that source directory leads to sporadic ctest errors - sometime in absolutely different places

etejedor:
> Nodes are not really full, but just huge number of created files in that source directory leads to sporadic ctest errors - sometime in absolutely different places

At least for what concerns the JupyROOT tests, once they will run in the build directory they will not be affected by the temporaries that were left in the source directory anymore. Were you thinking of some other tests that can be affected by the temporaries in the source dir?

etejedor:
> @etejedor, I guess same changes should be back-ported to 6.18, 6.20 and may be 6.16 branches
Otherwise nightly tests will continue pollute source dirs

Yes I will take care of that!

PR #522, opened 2020-05-20
Try to fix multicore test dependency

linev:
As dependency, one should provide real tests names
Also add dependency on ROOT libs when generate dictionary


linev:
@phsft-bot build

linev:
@phsft-bot build

PR #526, opened 2020-05-27
Adjust references for fascloning/execCheckClusterRange

oshadura:


oshadura:
CI seems to be not working for 6.16, merging it anyway.

PR #530, opened 2020-06-03
[PyROOT] Add read/write test for RooDataSet.

hageboeck:
This tests ROOT-10810.

### DIFF
--- a/python/pickle/PyROOT_writetests.py
+++ b/python/pickle/PyROOT_writetests.py
@@ -90,6 +90,25 @@ class PickleWritingSimpleObjectsTestCase( MyTestCase ):
       pickle.dump(  o, self.out1, protocol = 2 )
       cPickle.dump( o, self.out2, protocol = 2 )
 
+   def test6WriteCustomTypes( self ):
+      """[ROOT-10810] Test writing a RooDataSet with weights"""
+      # Silence the RooFit banner
+      ROOT.gEnv.SetValue("RooFit.Banner", 0)
+
+      var = ROOT.RooRealVar('var' ,'variable',0,10)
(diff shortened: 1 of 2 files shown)


python/pickle/PyROOT_writetests.py
```diff
@@ -90,6 +90,25 @@ def test5WriteCustomTypes( self ):
...
+      vs = ROOT.RooArgSet ( var , w )
+      ds = ROOT.RooDataSet('data', '', vs, ROOT.RooFit.WeightVar(w))
+
+      # Only fails with tree storage
+      ds.convertToTreeStore()
+      for i in range ( 10 ) :
+           var.setVal(i)
+           ds.add(vs, i+0.1)
+
+      pickle.dump(ds, self.out1, protocol = 2) ## <--- segmentation fault
```
etejedor, line 109: `segmentation fault` --> well not anymore :smile:

python/pickle/PyROOT_writetests.py
```diff
@@ -90,6 +90,25 @@ def test5WriteCustomTypes( self ):
       pickle.dump(  o, self.out1, protocol = 2 )
       cPickle.dump( o, self.out2, protocol = 2 )
 
+   def test6WriteCustomTypes( self ):
+      """Test writing a RooDataSet with weights"""
```
etejedor, line 94: If there is a JIRA associated with this, could you also add in this comment the number?

python/pickle/PyROOT_writetests.py
hageboeck: It's OK to leave it in, isn't it? In that case, people at least know where the problem showed.

PR #534, opened 2020-06-04
Change AddLinkedLibs with Load in regression-regression test

maxgalli:
goes with https://github.com/root-project/root/pull/5643


etejedor:
Same comment as for https://github.com/root-project/root/pull/5643. Merging to check the nightlies in master today and, if all goes well, merge into 6.22 tomorrow.

PR #535, opened 2020-06-04
Rename tclass_methods test executable to avoid build system race condition.

pcanal:
With the old name the dictionary source file tclass_methods.cpp was generated both by the test executable and the library
target.

We have:

```
ROOTTEST_GENERATE_EXECUTABLE(tclass_methods tclass_methods.cpp LIBRARIES Core Thread)

ROOTTEST_GENERATE_REFLEX_DICTIONARY(tclass_methods tclass_methods.h  SELECTION tclass_methods_selection.xml)

ROOTTEST_ADD_TEST(tclass_methods
                  EXEC ${CMAKE_CURRENT_BINARY_DIR}/tclass_methods
                  DEPENDS ${GENERATE_EXECUTABLE_TEST} ${GENERATE_REFLEX_TEST})
```

and we have ```ROOTTEST_GENERATE_REFLEX_DICTIONARY``` which calls ```REFLEX_GENERATE_DICTIONARY```

And  ```REFLEX_GENERATE_DICTIONARY ``` is defined in ```RootMacros.cmake``` and line 196 we have:

```
IF(TARGET ${dictionary})
    target_sources(${dictionary} PRIVATE ${gensrcdict})
  ENDIF()
```
so because of the ```ROOTTEST_GENERATE_REFLEX_DICTIONARY(tclass_methods....``` and the ```ROOTTEST_GENERATE_EXECUTABLE(tclass_methods ...```
We end up with ```tclass_methods (the executable)``` depending ont ```class_methods.cxx``` (as an irrelevant aside there is also a ```class_methods.cpp```)

The consequence is that the ```genreflex``` rule ends up in the ```Makefile``` for the executable ....

In addition ```ROOTTEST_GENERATE_REFLEX_DICTIONARY```creates 2 targets ```tclass_methods-libgen``` and ```tclass_methods-dictgen```
The makefiles for both contains the ```genreflex``` calls (both depends on ```tclass_methods.cxx```)

It looks like the designer intended for either ```tclass_methods-libgen``` or ```tclass_methods-dictgen``` to be used and not both.  And indeed in roottest the ```\*-dictgen``` appears to have a ```Makefile``` that is never called (i.e ```\*-dictgen``` is never actually used)

So there are 3 ```genreflex``` calls: for the ```executable, libgen and dictgen.```

Last we have:

```
DEPENDS ${GENERATE_EXECUTABLE_TEST} ${GENERATE_REFLEX_TEST})
```
which means that the test itself depends on the executable and on the ```libgen``` (and not the ```dictgen```)

and it *\*does\** mean that the dictionary generation can run in parallel .... since neither the ```tclass_method-libgen``` nor the ```executable``` depends on the others.


pcanal:
@phsft-bot build

PR #546, opened 2020-06-09
Do not use iomanip from TTree::MakeProxy in treeproxy test

linev:
Problem appears while `<iomanip>` is not provided by used ROOT includes
I just used directly `ostream::precision` method.
Is there way to add extra includes in code generated by `TTree::MakeProxy()`


linev:
[Here](https://lcgapp-services.cern.ch/root-jenkins/job/root-incremental-master/LABEL=ROOT-debian10-i386,SPEC=cxx14/8751/testReport/junit/projectroot.roottest.root/treeproxy/roottest_root_treeproxy_make/) is example of error during test

linev:
@Axel-Naumann Can you check this simple fix - otherwise we will have everywhere failing tests because of missing `<iomanip>` include

PR #554, opened 2020-06-16
Fix roottest on windows

bellenot:


bellenot:
The failures on `ROOT-performance-centos7-multicore/default` are unrelated

root/multicore/testSetAddressLoop.bat
```diff
@@ -0,0 +1,12 @@
...
+set COUNTER=0
+:loop
+echo "Run number %COUNTER%"
+testSetAddress.exe
+if errorlevel 1 (
+  echo "ERROR: The executable did not complete successfully!";
+  exit /b 1;
+)
+set /a COUNTER+=1
+if %COUNTER% LSS 33 goto loop
```
bernhardmgruber, line 12: Are batch files still a thing? We have powershell for years!

CMakeLists.txt
```diff
@@ -21,73 +22,118 @@ endif()
...
-  # root-config --prefix helps to retrieve ROOTSYS env variable.
-  # ROOT_BINDIR is set by ROOTConfig.cmake
-  find_program(ROOT_CONFIG_EXECUTABLE root-config PATHS ${ROOT_BINDIR})
-  if(NOT ROOT_CONFIG_EXECUTABLE)
-    message(ERROR "root-config is not found, please rerun configuration step.")
+if(MSVC)
+
+  set(libsuffix .dll)
+  set(exesuffix .exe)
+  set(CMAKE_CXX_FLAGS "/nologo /Zc:__cplusplus /MD /GR /EHsc- /W3 /D_WIN32")
```
bernhardmgruber, line 29: It is very surprising to me that we need to specify all these flags.

Maybe we should add a comment what they do.
I think /nologo should not be needed.
/Zc:__cplusplus ties the version reported by the equally named macro to the /std:c++xx argument.
/MD should be turned on by default by CMake, so I do not understand why it is necessary.
/GR enables RTTI and CMake should turn it on by default as well.
/EHsc- disables exceptions IIRC?
/W3 is the default for the warning level, I think we should not need it.
/D_WIN32 is defined by the compiler itself, so I wonder why you specify it.

eguiraud:
I looked at commit "Fix (or disable) the multicore tests on Windows " and it looks good to me :+1:

hageboeck:
I looked at `Don't run vectorisedPDFs tests on Windows`, and that's 👍

root/multicore/testSetAddressLoop.bat
bellenot: Right, but there is only one batch script...

CMakeLists.txt
bellenot: We need the same flags than ROOT, otherwise it doesn't build correctly (ask MS)

root/multicore/testSetAddressLoop.bat
bellenot: And since cling (llvm/clang) still has known open bugs and suffers from incompatibilities wit MS runtime, we need to disable tests until ROOT/cling gets fixed

python/cmdLineUtils/CMakeLists.txt
```diff
@@ -20,351 +20,358 @@ ROOTTEST_ADD_TEST(SimplePattern3
                   OUTREF SimplePattern3.ref)
 #########################################################################
 
-set(TOOLS_PREFIX ${ROOTSYS}/bin)
+if(MSVC)
+    # the command line tools works fine in the Windows command prompt but
+    # not from CTest, so let's add Python.exe and the .py file extension
+    set(TOOLS_PREFIX ${PYTHON_EXECUTABLE} ${ROOTSYS}/bin)
```
etejedor, line 26: You should use `${PYTHON_EXECUTABLE_Development_Main}` instead of `${PYTHON_EXECUTABLE}`, since the former will always point to the higher Python for which a development package has been found, and therefore that is the one that PyROOT tests should run with (the command line utils also use PyROOT). If you use `${PYTHON_EXECUTABLE}`, it can happen that it points to a Python that does not have development files (just the interpreter), which is fine for ROOT, but not for PyROOT.

bellenot: @etejedor  ${PYTHON_EXECUTABLE_Development_Main} is somehow not set, so I had  to set it [there](https://github.com/root-project/roottest/pull/554/commits/0a2465917ab179b71c0c02b3da468b13e8da2a1e)

CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ if(MSVC)
         PATH=$ENV{PATH}
         PYTHONPATH=${ROOTTEST_ENV_PYTHONPATH}:$ENV{PYTHONPATH})
   endif()
+  set(PYTHON_EXECUTABLE_Development_Main ${PYTHON_EXECUTABLE})
```
etejedor, line 43: This overriding is not correct, for the reasons I explained in my other comment about `${PYTHON_EXECUTABLE_Development_Main}`.

bellenot: @etejedor OK, thanks, so I will try to debug CMake to understand why PYTHON_EXECUTABLE_Development_Main is not properly set...

python/cmdLineUtils/CMakeLists.txt
etejedor: If should always be set if CMake finds at least one Python installation that has the development package. It is set here:

https://github.com/root-project/root/blob/master/cmake/modules/SearchRootCoreDeps.cmake

If it is not set we should find out why, all the Python tests should run with the Python that `${PYTHON_EXECUTABLE_Development_Main}` points to.

bellenot:
> I find the general changes to support Windows quite big. CMake should take off a lot of effort here. The sheer size of the PR tells me that we are probably not using CMake correctly to handle the platform specifics. I guess nobody has enough time to refactor the CMake scripts?

@bernhardmgruber developers *never* think about Windows (bash and configure/make are still required for old tests, which is why they are not run on Windows), and refactoring CMake scripts would be nice, if there is any volunteer...
 
> Furthermore, there are a lot `if(NOT MSVC OR win_broken_tests)`. I find this worrying. We should not disable tests, even if they are broken on a platform. I think this lets us forget about those tests. And these tests could just start to work with a newer compiler version or slight change in the code.

Cling (llvm/clang) still have known open bugs and suffers from incompatibilities wit MS runtime, we need to disable tests until ROOT/cling gets fixed. If you need all the technical details, I can give you some examples... And the purpose of the `win_broken_tests` flag is precisely to check if the disabled failing tests get fixed (we have the same in ROOT itself)

python/cmdLineUtils/CMakeLists.txt
bellenot: OK, thanks, I already found it in ROOT, but how is it set in roottest? Is it supposed to somehow inherit it?

eguiraud: ..and is it set if roottest is built independently of ROOT? :)

bellenot: > ..and is it set if roottest is built independently of ROOT? :)

What do you mean? `roottest` built without `ROOT`?

etejedor: @oshadura might know more

eguiraud: > > ..and is it set if roottest is built independently of ROOT? :)
> 
> What do you mean? `roottest` built without `ROOT`?

roottest built independently of root, against a pre-installed root

oshadura: Looks like we need to sync some part of https://github.com/root-project/root/blob/master/cmake/modules/SearchRootCoreDeps.cmake here https://github.com/root-project/roottest/blob/master/CMakeLists.txt#L136 to make standalone roottest working properly.

CMakeLists.txt
oshadura: @etejedor roottest.git missing all CMake part connected with standalone discovery of python.

python/cmdLineUtils/CMakeLists.txt
bellenot: > Looks like we need to sync some part of https://github.com/root-project/root/blob/master/cmake/modules/SearchRootCoreDeps.cmake here https://github.com/root-project/roottest/blob/master/CMakeLists.txt#L136 to make standalone roottest working properly.

Thanks @oshadura !

bellenot: > > > ..and is it set if roottest is built independently of ROOT? :)
> > 
> > 
> > What do you mean? `roottest` built without `ROOT`?
> 
> roottest built independently of root, against a pre-installed root

Yes, and that's the issue. It is now broken, as it's the way I tested the PR on Windows

bernhardmgruber:
> @bernhardmgruber developers _**never**_ think about Windows (bash and configure/make are still required for old tests, which is why they are not run on Windows), and refactoring CMake scripts would be nice, if there is any volunteer...

I think about Windows ;) Configure scripts should be dropped, but I guess you have either serious reasons or missing human resources. Therefore, something for the future probably.

> @bernhardmgruber Cling (llvm/clang) still have known open bugs and suffers from incompatibilities wit MS runtime, we need to disable tests until ROOT/cling gets fixed. If you need all the technical details, I can give you some examples... And the purpose of the `win_broken_tests` flag is precisely to check if the disabled failing tests get fixed (we have the same in ROOT itself)

I just checked what LLVM/clang 11 configure per default for Windows (master, a few weeks old). They have /MDd for debug builds and /MD for all other builds. But I have no big picture on how ROOT and cling use LLVM. So maybe you have a reason to use /MD for Debug builds as well.

So I guess you have regular builds with the broken tests enabled then? So this should be fine then.

bellenot:
> I think about Windows ;) Configure scripts should be dropped, but I guess you have either serious reasons or missing human resources. Therefore, something for the future probably.
>

We had a migration from `configure/make` to `CMake` several years ago, but people only tested on Linux/MacOS. And yes, could be a (not so interesting) project for the future...

> I just checked what LLVM/clang 11 configure per default for Windows (master, a few weeks old). They have /MDd for debug builds and /MD for all other builds. But I have no big picture on how ROOT and cling use LLVM. So maybe you have a reason to use /MD for Debug builds as well.
> 
> So I guess you have regular builds with the broken tests enabled then? So this should be fine then.

We are using a quite old version of LLVM/Clang, and we are planning to upgrade, hopefully still this year (Axel knows). and the /MDd issue is in ROOT only: it is to prevent having to build external dependencies of ROOT in debug mode too (long story), but it doesn't hurt, since we can still build in debug mode, we simply don't use the debug runtime dlls. And I personally try the broken tests when updating Visual Studio, it's how I discovered that with the latest version it's not possible to build llvm in debug mode anymore (Vassil is also looking into it, it looks like it's an issue with LLVM). In short: at every update of Visual Studio (every week or so) I check is something gets fixed, but most of the time, it breaks something in ROOT... (I also submitted a couple of bug reports - now fixed - at Microsoft).

CMakeLists.txt
bellenot: @etejedor @oshadura If you agree, I'll make a temporary fix, until the standalone version is properly fixed

bellenot:
@oshadura @etejedor I committed a change. Please let me know if we can go on,

CMakeLists.txt
```diff
@@ -165,7 +160,15 @@ endif()
 
 # Setup environment.
 set(ROOTTEST_ENV_PATH ${ROOT_BINDIR})
-set(ROOTTEST_ENV_PYTHONPATH ${ROOT_LIBRARY_DIR})
+if(MSVC)
+  if(NOT PYTHON_EXECUTABLE_Development_Main) # Should not happen!
+    # FIXME: Temporary workaround until standalone roottest is properly fixed
```
etejedor, line 165: If you would like to have this in to go on with your testing, then it is ok for me.

But note that even this workaround might not work, if `PYTHON_EXECUTABLE` as it was found by standalone roottest has a different value than what ROOT found for ${PYTHON_EXECUTABLE_Development_Main}`.

bellenot: Yes, I know, but good enough for now (since the standalone use is broken anyway). And I'll remove the workaround once this is properly fixed.

bellenot:
@Axel-Naumann @pcanal @oshadura @amadio does anyone wants to review (part of) this PR?

root/tree/readcin/CMakeLists.txt
```diff
@@ -1,4 +1,8 @@
-ROOTTEST_GENERATE_EXECUTABLE(readFromCin readFromCin.cxx LIBRARIES Core Hist RIO Net Graf Graf3d Gpad Tree Rint Postscript Matrix Physics MathCore Thread MultiProc)
+if(MSVC)
+    ROOTTEST_GENERATE_EXECUTABLE(readFromCin readFromCin.cxx LIBRARIES Core Hist RIO Net Graf Graf3d Gpad Tree Rint Postscript Matrix Physics MathCore Thread)
+else()
+    ROOTTEST_GENERATE_EXECUTABLE(readFromCin readFromCin.cxx LIBRARIES Core Hist RIO Net Graf Graf3d Gpad Tree Rint Postscript Matrix Physics MathCore Thread MultiProc)
```
pcanal, line 4: Would it be worth using a ```${OptionalMultiProc}``` variable in multiple place rather than ```if (MSCV)``` in those same places?

root/multicore/loadImt.cpp
```diff
@@ -1,6 +1,8 @@
 #include "TSystem.h"
 #include "TInterpreter.h"
+#ifndef _MSC_VER
```
pcanal, line 3: Did we used to use ```_WIN32``` (and ```R__WIN32```) for this purpose?

bellenot: Well, I prefer `_MSC_VER` it's more explicit for `Visual Studio` cases (there is also `clang` and `MingW gcc`)

root/tree/readcin/CMakeLists.txt
bellenot: I'll check, but `MultiProc` is definitively not supported on Windows (and I'm not going to do anything about it)

root/meta/naming/CMakeLists.txt
```diff
@@ -1 +1,12 @@
-ROOTTEST_ADD_AUTOMACROS(DEPENDS namingMatches.cxx)
+ROOTTEST_ADD_AUTOMACROS(DEPENDS namingMatches.cxx
+                        EXCLUDE execCheckNaming)
+
+if(MSVC)
+    set(EXECCHECKNAMING_OUTREF execCheckNaming_win32.ref)
```
pcanal, line 5: If I remember correctly we can auto detect the sizeof(long) and adjust the name accordingly?  (See the use of ROOTBITS in cmake/modules/RoottestMacros.cmake; eg for example is _win32 if th file exist and the platform is windows).

cling/dict/ROOT-7879/CMakeLists.txt
```diff
@@ -1,12 +1,19 @@
 # ROOT-7879
 ROOTTEST_ADD_TEST(skipLinkDefDict
-                  COMMAND ${ROOT_rootcling_CMD} -f libSkippingDict.cxx -rmf libSkipping.rootmap -rml libSkipping.so  ${CMAKE_CURRENT_SOURCE_DIR}/skippingHeader.h ${CMAKE_CURRENT_SOURCE_DIR}/skipLinkDef.h
+                  COMMAND ${ROOT_rootcling_CMD} -f libSkippingDict.cxx -rmf libSkipping.rootmap -rml libSkipping${libsuffix} ${CMAKE_CURRENT_SOURCE_DIR}/skippingHeader.h ${CMAKE_CURRENT_SOURCE_DIR}/skipLinkDef.h
                   OUTREF skipLinkDefDict.ref)
 
+if(MSVC)
+    set(grep_cmd "findstr")
```
pcanal, line 7: This is used twice, could it be factor out in the cmake/* files.?

root/meta/naming/CMakeLists.txt
bellenot: It's not only a 32 bit issue, in this case the symbols (mangled names) are also different between Windows and Linux

cling/dict/ROOT-7879/CMakeLists.txt
bellenot: Yes, but we also have grep with different arguments... Anyway, I'll move it elsewhere

bellenot: Done

root/meta/naming/CMakeLists.txt
bellenot: I'll check how to test if a file exists in CMake. IIUC you suggest to test if it is Windows, if it is 32 bit, and if the ref file exists?

pcanal: Sure (i.e. the same checks you are doing now, just in one single place).

root/multicore/loadImt.cpp
pcanal: Fair enough ....

root/tree/readcin/CMakeLists.txt
pcanal: Humm what I am proposing:
```
ROOTTEST_GENERATE_EXECUTABLE(readFromCin readFromCin.cxx LIBRARIES Core Hist RIO Net Graf Graf3d Gpad Tree Rint Postscript Matrix Physics MathCore Thread ${OptionalMultiProc} )
```
with ${OptionalMultiProc} being a ```empty``` on windows and ```MultiProc``` everywhere else.

bellenot: OK, will do that, thanks for the suggestion

bellenot: Done

root/meta/naming/CMakeLists.txt
bellenot: Done

pcanal: I am a bit confused.   On my view of the page I still see a lot of conditional setting of reference file names .. what am I missing?

bellenot: They are outdated. See https://github.com/root-project/roottest/pull/554/commits/b3a16c5c8e7b1d9bcd8f50a2aa195759c886fe68

bellenot: I.e. there is a small yellow `Outdated` box after the file name `root/meta/naming/CMakeLists.txt` `Outdated`

bellenot:
Errors are unrelated. Merging. Thanks all for the reviews

PR #555, opened 2020-06-23
Fix (or disable) remaining failg tests on Windows

bellenot:
Following the latest changes in ROOT, some test are now working (then remove the WILLFAIL flag), and mark the forgotten failing ones as WILLFAIL

### DIFF
--- a/root/dataframe/CMakeLists.txt
+++ b/root/dataframe/CMakeLists.txt
@@ -24,12 +24,14 @@ ROOTTEST_ADD_TEST(test_trainCache
 # ROOT-9628
 ROOTTEST_ADD_TEST(test_chainZombieFile
                   COPY_TO_BUILDDIR input1.root input2.root
+                  ${MIGHT_FAIL}
                   MACRO test_chainZombieFile.C)
 
 # ROOT-9366
 # We use this to write something like the FCC data model, dictionaries done with aclic
 ROOTTEST_ADD_TEST(writeFcc
-                  MACRO writeFcc.C+)
+                  MACRO writeFcc.C+
+                  ${MIGHT_FAIL})
 
 # We build and run an executable that reads the file created above without dictionaries
 ROOTTEST_GENERATE_EXECUTABLE(test_readFcc readFcc.C LIBRARIES ${DFLIBRARIES})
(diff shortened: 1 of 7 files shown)


root/dataframe/test_trainCache.C
```diff
@@ -1,3 +1,5 @@
+R__LOAD_LIBRARY(libROOTDataFrame);
```
eguiraud, line 1: why is this required? or: why was it not required before?

root/dataframe/CMakeLists.txt
```diff
@@ -24,19 +24,22 @@ ROOTTEST_ADD_TEST(test_trainCache
 # ROOT-9628
 ROOTTEST_ADD_TEST(test_chainZombieFile
                   COPY_TO_BUILDDIR input1.root input2.root
+                  ${MIGHT_FAIL}
```
eguiraud, line 27: the variable name is misleading, should really be called `WILL_FAIL_ON_WIN` or similar

root/dataframe/test_trainCache.C
bellenot: I don't know, but without that it fails on Windows...

root/dataframe/CMakeLists.txt
bellenot: It's already user all over the place, but I'll change them all...

eguiraud: thanks. it's terrible to have most RDF tests marked "MIGHT_FAIL", it's slightly less terrible if they are marked "WILL_FAIL_ON_WINDOWS" :)

root/dataframe/test_trainCache.C
eguiraud: alright, can you please enclose it in an `#ifdef WINDOWS`, if nothing else for documentation purposes?

eguiraud: (rationale: we want to keep testing that autoloading works on other platforms)

bellenot: Fair enough, will do

root/dataframe/CMakeLists.txt
bellenot: According to Axel, it's a known issue in llvm/clang/cling

root/dataframe/test_trainCache.C
bellenot: Done

bellenot:
> Thanks!

Thanks to you!

bellenot:
Errors are unrelated - merging

PR #565, opened 2020-07-27
[ntuple] add rntuple merge test skeleton

mxxo:
Add empty unit test intended to test RNTuple merges. We add a new macro `execRNTupleMerge.C` which will create RNTuples on the fly to be merged (following what happens for `TTree`, etc. in `execFileMerger.C`) 

We can add future RNTuple merge tests using the files created as part of execRNTupleMerge (e.g. testing for hadd): 

```cmake 
ROOTTEST_ADD_TEST(hadd-rntuple
                     COMMAND ${ROOT_hadd_CMD} -f merged.root ntuple1.root ntuple2.root
                     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
                     DEPENDS execRNTupleMerge)
```

### DIFF
--- a/root/io/filemerger/CMakeLists.txt
+++ b/root/io/filemerger/CMakeLists.txt
@@ -10,6 +10,10 @@ ROOTTEST_ADD_TEST(execFileMerger
                   MACRO execFileMerger.C
                   OUTREF references/execFileMerger.ref)
 
+ROOTTEST_ADD_TEST(execRNTupleMerge
+                  MACRO execRNTupleMerge.C
+                  OUTREF references/execRNTupleMerge.ref)
+
 # FIXME: Should be CTEST fixtures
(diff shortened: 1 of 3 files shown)


mxxo:
Closing in favour of a new RNTuple tutorial about merging.

pcanal:
The tutorials and the test are not mutual exclusive.  A tutorials should have code that looks the way user code should look like.  While the test code be more pedantic about testing and checking.  For example the test code might/should check that the resulting compression and size are as expected, the tutorials should not.

mxxo:
That's a very good point, I am happy to include them both, especially since they will likely share a lot of code.

Axel-Naumann:
@jblomer what are the plans with this PR?

guitargeek:
Seems outdated to me

PR #569, opened 2020-08-08
Adjust reference file.

vgvassilev:
This change follows root-project/root#5847

### DIFF
--- a/root/rint/TabCom_input.txt
+++ b/root/rint/TabCom_input.txt
@@ -1,4 +1,3 @@
-TObj	
 TObject::k	
 gROOT->GetListOfT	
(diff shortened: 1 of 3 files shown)


root/rint/TabCom_input.txt
```diff
@@ -1,4 +1,3 @@
-TObj
```
pcanal, line 1: Why was that line useful/interesting before and is no longer of interest?

vgvassilev: That moved in the TTabComTests.cxx in the ROOT repository because the return results started depending on things like ROOT7 for TObjHolder and so on. That is why I decided to sink that code there to be able to check better for configuration.

PR #572, opened 2020-08-13
TClingMethodInfo now skips deleted functions:

Axel-Naumann:
Update ref files, now that TClass printing / TMethod iteration skips
deleted functions. This happens e.g. for const members, where the
compiler-generated op= is deleted.


Axel-Naumann:
Tests https://github.com/root-project/root/pull/6173

PR #586, opened 2020-08-26
[PyROOT] Check if CPPYY_BACKEND_LIBRARY points to an existing file

etejedor:
If not, unset the environment variable. This will give to cppyy
the opportunity to find the library on its own, by looking in its
default directories.

This should solve this kind of failures in the conda builds:
https://lcgapp-services.cern.ch/root-jenkins/view/conda/job/conda-nightlies/104/testReport/projectroot.python/cling/roottest_python_cling_api/

which are due to the fact that the tests are run in standalone
mode and, therefore, {CMAKE_BINARY_DIR}/lib does not contain
the ROOT libraries.

### DIFF
--- a/python/cling/cppyy_backend_check.h
+++ b/python/cling/cppyy_backend_check.h
@@ -1,0 +1,35 @@
+using std::string;
+
+std::pair<string,string> SplitPathAndName(const char *path)
+{
+   char sep = '/';
+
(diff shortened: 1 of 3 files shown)


eguiraud (approved):
LGTM

etejedor:
Failures are unrelated (and already reported).

PR #592, opened 2020-09-11
[DF] Fix test for ROOT-9674

eguiraud:
Column type is RVec<int>, not int, and RDF gets it wrong.
https://github.com/root-project/root/pull/6258 fixes the type inference
problem (and therefore this test) for master.


pcanal (approved):

pcanal:
@phsft-bot build only on ROOT-performance-centos7-multicore/default

pcanal:
@phsft-bot build just on ROOT-performance-centos7-multicore/default

PR #595, opened 2020-09-15
[PyROOT] Add test for using function decl

stwunsch:
Reported on mattermost that "import ROOT; ROOT.TGraphMultiErrors().SetLineColor(0)"
breaks with 6.22 and is now fixed in master due to the ROOT PR #6277.

### DIFF
--- a/python/cpp/PyROOT_cpptests.py
+++ b/python/cpp/PyROOT_cpptests.py
@@ -17,7 +17,8 @@ from functools import partial
 
 __all__ = [
    'Cpp1LanguageFeatureTestCase',
-   'Cpp2ClassNamingTestCase'
+   'Cpp2ClassNamingTestCase',
+   'Cpp3UsingDeclarations',
 ]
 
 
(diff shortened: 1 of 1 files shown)


stwunsch:
@etejedor Is this the correct place for the test?

etejedor (changes requested):
Thanks @stwunsch ! I wonder if we could create a new class of tests for using declarations here:

https://github.com/root-project/roottest/blob/master/python/cpp/PyROOT_cpptests.py

since this is likely not the only test we will want to add for using declarations (there are several tickets related to it).

stwunsch:
> Thanks @stwunsch ! I wonder if we could create a new class of tests for using declarations here:
> 
> https://github.com/root-project/roottest/blob/master/python/cpp/PyROOT_cpptests.py
> 
> since this is likely not the only test we will want to add for using declarations (there are several tickets related to it).

I've moved the test and introduced a new class of tests related to using declarations!

etejedor (approved):

PR #596, opened 2020-09-23
[PyROOT] Add more tests for using declarations and Python-C++ inheritance

etejedor:
Extracted from the following JIRA issues:
https://sft.its.cern.ch/jira/browse/ROOT-10872
https://sft.its.cern.ch/jira/browse/ROOT-11010
https://sft.its.cern.ch/jira/browse/ROOT-10789
https://sft.its.cern.ch/jira/browse/ROOT-10786


stwunsch (approved):
LGTM as it is! However, would it make sense to also test that the derived (protected) methods also work when they are called? I could image that there are also issues hidden.

etejedor:
> However, would it make sense to also test that the derived (protected) methods also work when they are called? I could image that there are also issues hidden.

Done, thanks @stwunsch !

PR #600, opened 2020-09-26
[skip-ci] Test that we correctly avoid auto-parsing when we can/should

pcanal:


pcanal:
See https://github.com/root-project/root/pull/6445
and https://github.com/root-project/root/issues/6443

PR #606, opened 2020-10-07
Add a generic executor test and improve the current executor tests

xvallspl:
Goes together with ROOT's PR https://github.com/root-project/root/pull/6327

I expect these tests to be ported to gtest eventually, but it's not straightforward. Let's keep them here for the moment.


xvallspl:
@phsft-bot build

xvallspl:
@phsft-bot  build!

PR #607, opened 2020-10-07
[PyROOT][6467] Add test for inheritance from TChain

etejedor:
Test for https://github.com/root-project/root/pull/6582


stwunsch (approved):
Nice! :)

PR #612, opened 2020-10-13
[PyROOT][6578] Add test for constructor with all default parameters

etejedor:
Test for https://github.com/root-project/root/issues/6578 on the Python side.


vepadulano (approved):
LGTM, the original issue is already fixed in cling

PR #635, opened 2021-01-12
[PyROOT][6470] Tests for cross inheritance

etejedor:
Tests for https://github.com/root-project/root/pull/7022

### DIFF
--- a/python/regression/PyROOT_regressiontests.py
+++ b/python/regression/PyROOT_regressiontests.py
@@ -798,6 +798,40 @@ class Regression24CppPythonInheritance(MyTestCase):
        self.assertEqual(a.bar(), 3)
        self.assertRaises(TypeError, a.bar, 0)
 
+   def test10DeepHierarchyVirtualCall(self):
+       """Virtual call resolution in deep hierarchy (C++->Py->Py)"""
+       # 6470
+
+       cppyy.gbl.gInterpreter.Declare('''
+       class CppBase6470 {
+       public:
+          virtual ~CppBase6470() {}
+          virtual int fun1() const { return 1; }
+          virtual int fun2() const { return fun1(); }
+       };
+       ''')
+
+       CppBase = cppyy.gbl.CppBase6470
+
+       class PyA(CppBase):
+          def fun1(self): return 11
+
+       class PyB(PyA):
+          def fun1(self): return 111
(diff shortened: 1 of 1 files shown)


python/regression/PyROOT_regressiontests.py
```diff
@@ -798,6 +798,65 @@ class MyPyClass6345(cppyy.gbl.MyClass6345):
...
+       cppyy.gbl.gInterpreter.Declare('''
+       class MyCppClass11 {
+       public:
+          int foo() { return 3; }
+          int bar() { return 4; }
+          virtual ~MyCppClass11() {}
+       };
+       ''')
+
+       # Multiple Python classes and just one C++ class are supported
```
stwunsch, line 853: Do we also need a test for C++ -> C++ -> Python? Am I right that this is trivially handled on the C++ side?

python/regression/PyROOT_regressiontests.py
```diff
@@ -798,6 +798,65 @@ class MyPyClass6345(cppyy.gbl.MyClass6345):
...
+       class PurePy1:
+          def foo(self): return 1
+
+       class PurePy2:
+          def bar(self): return 2
+
+       cppyy.gbl.gInterpreter.Declare('''
+       class MyCppClass11 {
+       public:
+          int foo() { return 3; }
```
stwunsch, line 847: What about protected methods? Do they also work as expected? I vaguely remember some discussion here (?)

stwunsch (approved):
LGTM! Some minor questions :)

python/regression/PyROOT_regressiontests.py
etejedor: Yes, that works too, it's tested here:
https://github.com/root-project/roottest/blob/master/python/regression/PyROOT_regressiontests.py#L643

Here it does not really matter if it's public or protected, in both cases the MRO is respected (I quickly tried to be 100% sure :D).

python/regression/PyROOT_regressiontests.py
etejedor: Yes you are right, that would be handled on the C++ side. It's the construction of the deep hierarchy on the Python side that was not working properly because the inheritance was not "chained" correctly.

PR #636, opened 2021-01-14
Disable tests requiring MultiProc (not supported on Windows)

bellenot:


xvallspl:
This test should be run on Windows too, I will think about it

bellenot:
Well, then I'll check how I can only disable the Multiproc part

bellenot:
superseded by #637

PR #637, opened 2021-01-14
fix tExecutor test on Windows

xvallspl:


xvallspl:
@bellenot Would this be enough? to deal with https://github.com/root-project/roottest/pull/636

bellenot:
> @bellenot Would this be enough? to deal with #636

I think so (it's exactly what I wanted to try). I'll approve it as soon as I managed to successfully run it on Windows

root/multicore/CMakeLists.txt
```diff
@@ -87,7 +87,11 @@ if(ROOT_imt_FOUND)
                     DEPENDS ${GENERATE_EXECUTABLE_TEST})
 endif()
 
-ROOTTEST_GENERATE_EXECUTABLE(tExecutor tExecutor.cxx LIBRARIES Core Imt MultiProc Net TreePlayer RIO Hist MathCore)
+if(NOT MSVC OR win_broken_tests)
```
bellenot, line 90: ``

root/multicore/tExecutor.cxx
```diff
@@ -25,10 +25,12 @@ int main() {
    }
 #endif
 
+#ifndef R__WIN32
```
bellenot, line 28:
```suggestion
#ifndef _MSC_VER
```

root/multicore/tExecutor.cxx
```diff
@@ -43,8 +45,10 @@ int main() {
    res = ExecutorTest(ex1);
    if(res) return res+offset+4;
 #endif
+#ifndef R__WIN32
```
bellenot, line 48:
```suggestion
#ifndef _MSC_VER
```

bellenot (changes requested):
In general `_MSC_VER` is preferred over `R__WIN32` which can also be defined when using different compilers supporting more features than MSVC

bellenot (approved):
Thanks!

PR #639, opened 2021-01-15
cmdLineUtils: survive the input being already created.

pcanal:
rootcp (unlike /bin/cp and as intended) is not copying the raw file but copy the content from the input to the output
It defaults to 'appending' mode rather than over-write/replace and thus if the file already exists (for example after
an interupted run of ctest) then the file use for the test contains more keys than expected.

This fixes https://github.com/root-project/root/issues/7054


etejedor (approved):
LGTM thanks!

PR #640, opened 2021-01-20
Fix out-of-range access in executorTests

hahnjo:
Both `htot` and `hred` have 10 bins, so including the bins for under- and overflow the maximum index to `GetBinContent` should be 11.


hahnjo:
@phsft-bot build just on mac1014/python3, mac11.0/cxx17

hahnjo:
Actually, `TH1::GetBinContent` is guarded against invalid accesses (https://github.com/root-project/root/blob/master/hist/hist/src/TH1.cxx#L4995), but I think the change is correct.

xvallspl (approved):

PR #642, opened 2021-01-21
Fix parallel execution on Windows

bellenot:
This partially fixes the parallel execution on Windows. The `--repeat until-pass` option is also required. For example: `ctest -j8 --repeat until-pass:8 -C Release`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,7 @@ if(MSVC)
 
   if(NOT win_broken_tests)
     set(WILLFAIL_ON_WIN32 WILLFAIL)
+    set(RUN_SERIAL_ON_WIN32 RUN_SERIAL)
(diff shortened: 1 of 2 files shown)


eguiraud:
why can't these tests run concurrently?

bellenot:
> why can't these tests run concurrently?

I don't know, maybe memory limitation (32 bit), or resource sharing... Do you want me to debug and find out?

eguiraud:
yes please: this is probably the right patch if the problem is memory usage (the alternative is making our unit tests less memory-hungry). but if it's something else, the right fix might be different.

bellenot:
@eguiraud OK, then it will take quite some time, I cannot build/run the tests in debug mode on Windows for the time being. I first have to fix CMake.

PR #646, opened 2021-01-28
Fix several CI failures on Windows

bellenot:
- use the proper `$<CONFIG>` value instead of `Release`
- disable several tests which are failing because of their PATH being too long (>260 characters) when running in Jenkins


Axel-Naumann (approved):

bellenot:
Merging (failures are unrelated)

PR #648, opened 2021-01-29
Test TExecutor+{MT,MP} separately

eguiraud:
Using TProcessExecutor in a multi-thread program is problematic
(because calling fork() from a multi-thread program is).

So let's make sure to not mix multi-threading and multi-processing
with TExecutor in the same executable.


root/multicore/CMakeLists.txt
```diff
@@ -87,15 +87,17 @@ if(ROOT_imt_FOUND)
...
-
-ROOTTEST_GENERATE_EXECUTABLE(tExecutor tExecutor.cxx LIBRARIES Core Imt ${MULTIPROC_LIB} Net TreePlayer RIO Hist MathCore)
+ROOTTEST_GENERATE_EXECUTABLE(tExecutorMT tExecutorMT.cxx LIBRARIES Core Imt Net TreePlayer RIO Hist MathCore)
+ROOTTEST_ADD_TEST(tExecutorMT
+                  EXEC ${CMAKE_CURRENT_BINARY_DIR}/tExecutorMT
+                  DEPENDS ${GENERATE_EXECUTABLE_TEST})
 
-ROOTTEST_ADD_TEST(tExecutor
-                  EXEC ${CMAKE_CURRENT_BINARY_DIR}/tExecutor
+if(NOT MSVG) # TProcessExecutor is not available on windows
```
hahnjo, line 95: `MSVC`? Not sure if this meant that the test was always executed or never...

root/multicore/tExecutorMT.cxx
```diff
@@ -0,0 +1,49 @@
...
+#ifdef R__USE_IMT
+   res = ExecutorTest(ex1);
+   if (res)
+      return res + offset + 4;
+#endif
+   res = ExecutorTest(ex2);
+   if (res)
+      return res + 2 * offset + 4;
+
+   return res ? res + 3 * offset + 4 : 0;
```
hahnjo, line 48: Just `return 0`, `res` was already checked in the previous `if` (was unnecessary in the original test already)

hahnjo (changes requested):

root/multicore/CMakeLists.txt
eguiraud: damn, thanks xD

we want to never execute it on windows -- there is no `fork` there, so no `TProcessExecutor`

hahnjo: I get that, but what does the typo do? I would have thought that it makes the test be executed always since `MSVG` is not defined, but Jenkins didn't report an error :thinking:

hahnjo: Ah, the PR build doesn't (yet) enable `roottest` on Windows. Never mind then.

hahnjo (approved):

PR #649, opened 2021-02-02
Set the CMake RUN_SERIAL option for ACLiC on Windows

bellenot:


bellenot:
Failure are unrelated. Merging

PR #650, opened 2021-02-02
[MultiProc] Fix library deps in tExecutorMT test

eguiraud:
Even if TProcessExecutor is not used at runtime, TExecutor still
depends on the libMultiProc (what contains TProcessExecutor) at
link time.

This should fix a linking failure for this test in the conda builds (failure is [here](https://lcgapp-services.cern.ch/root-jenkins/job/conda-nightlies/313/testReport/junit/projectroot.root/multicore/roottest_root_multicore_tExecutorMT_build/)).

### DIFF
--- a/root/multicore/CMakeLists.txt
+++ b/root/multicore/CMakeLists.txt
@@ -87,17 +87,22 @@ if(ROOT_imt_FOUND)
                     DEPENDS ${GENERATE_EXECUTABLE_TEST})
 endif()
 
-ROOTTEST_GENERATE_EXECUTABLE(tExecutorMT tExecutorMT.cxx LIBRARIES Core Imt Net TreePlayer RIO Hist MathCore)
+if(NOT MSVC)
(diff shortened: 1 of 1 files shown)


root/multicore/CMakeLists.txt
```diff
@@ -87,18 +87,23 @@ if(ROOT_imt_FOUND)
                     DEPENDS ${GENERATE_EXECUTABLE_TEST})
 endif()
 
-ROOTTEST_GENERATE_EXECUTABLE(tExecutorMT tExecutorMT.cxx LIBRARIES Core Imt Net TreePlayer RIO Hist MathCore)
+if(NOT MSVC)
+   ROOTTEST_GENERATE_EXECUTABLE(tExecutorMP tExecutorMP.cxx LIBRARIES Core MultiProc Imt Net TreePlayer RIO Hist MathCore)
```
pcanal, line 91: It is a bit surprising that the MultiProc dependent is not coming transitively from the dependency on Imt ....

eguiraud: I think it's the usual problem with roottest built independently: the target that brings dependencies with it would be `ROOT::MultiProc`. Unfortunately that target does not exist when roottest is built as part of ROOT.

The only solution I know of is to list all dependencies explicitly.

PR #652, opened 2021-02-03
Prefer ROOT::Lib to Lib when linking some tests

eguiraud:
The new spelling ensures transitive dependencies are brought in.


eguiraud:
@phsft-bot build please

PR #656, opened 2021-02-05
[MultiProc] Set ROOT to batch mode before using TProcessExecutor

eguiraud:
On MacOS, Cocoa spawns threads. That's very bad for TProcessExecutor's `fork`:
forking a multi-thread program can easily break, see e.g. `man 2 fork`.
Setting ROOT to batch mode works around the issue.

I had missed this test in the previous pass :/

### DIFF
--- a/root/multicore/tExecutorMP.cxx
+++ b/root/multicore/tExecutorMP.cxx
@@ -10,5 +10,8 @@ int main()
       return 1;
    }
 
+   // On MacOS, Cocoa spawns threads. That's very bad for TProcessExecutor's `fork`:
+   // forking a multi-thread program can easily break, see e.g. `man 2 fork`.
+   gROOT->SetBatch(kTRUE);
    return ExecutorTest(ex);
 }


Axel-Naumann (approved):

root/multicore/tExecutorMP.cxx
```diff
@@ -10,5 +10,8 @@ int main()
       return 1;
    }
 
+   // On MacOS, Cocoa spawns threads. That's very bad for TProcessExecutor's `fork`:
+   // forking a multi-thread program can easily break, see e.g. `man 2 fork`.
+   gROOT->SetBatch(kTRUE);
```
hahnjo, line 15: Is that early enough or should this go before creating the `TExecutor` (which might initialize ROOT)? I was trying to check where the threads are spawned, but `macphsft19` says `error: process exited with status -1 (developer mode is not enabled on this machine and this is a non-interactive debug session.)` when trying to launch the executable in `lldb` :neutral_face:

hahnjo: I think some limited success on `macphsft20` tells me that actually the first call to `pthread_create` happens after the first `fork`, which makes sense because the dispatching might only be needed for communicating with the spawned threads. (Note that future `fork`s are still a problem then). So we're probably good with calling `SetBatch` here.

hahnjo (approved):

eguiraud:
Thanks for checking @hahnjo , fair question about when exactly the threads are spawned :+1:

PR #657, opened 2021-02-11
Set the RUN_SERIAL option when building tests on Windows

bellenot:


bellenot:
The failure is unrelated. Merging.

PR #661, opened 2021-02-19
Fix several issues with roottest on Windows

bellenot:
- delete several `.rootmap` files to prevent autoloading leading to this kind of error:
      Warning in <TInterpreter::ReadRootmapFile>: class  Top found in runSplitMismatch_C is already in abstract_C
      ./abstract.C:1:7: error: redefinition of 'Top'
   or this other kind:
      Error in <reuseMethod>: Found a dictionary for vector<TObject*>, state: 4
 - add several ROOTTEST_COMPILE_MACRO() to compile the macro beforehand, to prevent DLL access error

This allows to:
 - enable several previously disabled tests
 - remove several WILL_FAIL flags
 - might allow to run in parallel without the `--repeat until-pass:3` flag
 - allow to re-run roottest several times in a row (should fix incremental and PR builds)

### DIFF
--- a/root/io/stdarray/CMakeLists.txt
+++ b/root/io/stdarray/CMakeLists.txt
@@ -72,7 +72,8 @@ ROOTTEST_ADD_TEST(modelReadDict2TFile
                   MACRO  modelReadDict2.C
                   MACROARG "\"model.root\""
                   OUTREF modelReadDict2.ref
-                  DEPENDS aclicModelWrite)
+                  DEPENDS aclicModelWrite
+                  POSTCMD ${CMAKE_COMMAND} -E remove modelForReadDict2_cpp.rootmap)
 
 # ROOTTEST_ADD_TEST(modelReadDict2TXMLFile
 #                   MACRO  modelReadDict2.C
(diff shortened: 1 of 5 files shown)


bellenot:
@pcanal  I guess I should apply those changes for Windows only (with appropriate `if(MSVC)`)
(I created this PR mostly to test on the other platforms)

root/io/stdarray/CMakeLists.txt
```diff
@@ -72,7 +72,8 @@ ROOTTEST_ADD_TEST(modelReadDict2TFile
                   MACRO  modelReadDict2.C
                   MACROARG "\"model.root\""
                   OUTREF modelReadDict2.ref
-                  DEPENDS aclicModelWrite)
+                  DEPENDS aclicModelWrite
+                  POSTCMD ${CMAKE_COMMAND} -E remove modelForReadDict2_cpp.rootmap)
```
pcanal, line 76: That's dangerous and/or will lead to confusion.  If the test requires the modelForReadDict2_cpp.rootmap file and it is run a second time, it will (mysteriously) fail the 2nd time (since the library wont be rebuilt).

pcanal:
> delete several .rootmap files to prevent autoloading leading to this kind of error:

Rather than deleting those (see my other comment), we should consider disabling their generation for those directories (see for example commit 0357ba1a43 ./root/io/datamodelevolution/misc/.rootrc).

If some test in those directory actually requires the rootmap files (i.e. start failing on linux), we should move either the test that can't handle or the test that requires the rootmap file into a separate sub-directory where we can disable (or enable) the rootmap file generation.

bellenot:
@pcanal Thanks for the review and the comments. That was my first attempt to fix multiple runs of roottest without cleanup in between, which is leading to many issue on Windows. Note that it looks like that doesn't perturb the other platforms. Anyway, I will try what you suggest and continue working on it.

bellenot:
@pcanal FYI using the proposed `.rootrc` solution doesn't work. Still working on it...

bellenot:
@pcanal I think I found the solution (even possibly fixing other tests) thanks to you comment about `.rootrc`. I will most probably create another PR

PR #663, opened 2021-02-23
Fix several issues with roottest on Windows

bellenot:
Add `ACLiC.LinkLibs:  1` in the `.rootrc` files for Windows (thanks Philippe). According to [the documentation](https://github.com/root-project/root/blob/master/config/rootrc.in#L384-L390), this is already the default for the other platforms, while the default is `3` for Windows.

This allows to:
 - enable several previously disabled tests
 - remove several `WILL_FAIL` flags
 - might allow to run in parallel without the `--repeat until-pass:3` flag
 - re-run roottest several times in a row (should fix incremental and PR builds)

### DIFF
--- a/cmake/modules/RoottestMacros.cmake
+++ b/cmake/modules/RoottestMacros.cmake
@@ -49,6 +49,7 @@ function(ROOTTEST_ADD_TESTDIRS)
     # create .rootrc in binary directory to avoid filling $HOME/.root_hist
     file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/${d}/.rootrc "
 Rint.History:  .root_hist
+ACLiC.LinkLibs:  1
 ")
   endforeach()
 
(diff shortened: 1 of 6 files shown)


bellenot:
@phsft-bot  build with flags -DCTEST_TEST_EXCLUDE_NONE=ON

bellenot:
Test failures are obviously unrelated (since I only changed roottest...)

pcanal:
The stress failure and climate failure are indeed unrelated.

The failure in 'projectroot.roottest.root.meta.genreflex.ROOT-5594.roottest_root_meta_genreflex_ROOT_5594_5594' is "new to me" but then again that's on Windows :)

pcanal (approved):

bellenot:
> The stress failure and climate failure are indeed unrelated.
> 
> The failure in 'projectroot.roottest.root.meta.genreflex.ROOT-5594.roottest_root_meta_genreflex_ROOT_5594_5594' is "new to me" but then again that's on Windows :)

Thanks for the comment and yes, let's see how it goes from there on. Then I'll tackle the remaining ones causing problems 😉

PR #667, opened 2021-03-02
[DF] Improvements to CheckDefine

eguiraud:
Companion PR for https://github.com/root-project/root/pull/7333


eguiraud:
*sigh* roottest does not rebase branches on master before building..

PR #669, opened 2021-03-03
Fixes to tree/cloning/CMakeLists.txt

eguiraud:
* define variable before it is first used
* do not assume (root-config --tutdir) is $ROOTSYS/tutorials

### DIFF
--- a/root/tree/cloning/CMakeLists.txt
+++ b/root/tree/cloning/CMakeLists.txt
@@ -1,7 +1,9 @@
 if(ROOTTEST_DIR)
       set(ROOT_EVENT_DIR ${ROOTTEST_DIR}/root/treeformula/event/)
+      set(ROOT_DIR ${ROOTSYS})
 else()
       set(ROOT_EVENT_DIR ${ROOT_SOURCE_DIR}/roottest/root/treeformula/event/)
+      set(ROOT_DIR ${ROOT_SOURCE_DIR})
 endif()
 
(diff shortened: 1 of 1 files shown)


eguiraud:
Philippe mentioned that `root-config --tutdir` might be wrong when ROOT is ready to be installed somewhere else and we run the tests from the build directory.

And `hsimple.C` is use also in other tests and we should check that they can find it too.

root/tree/cloning/CMakeLists.txt
```diff
@@ -1,7 +1,11 @@
 if(ROOTTEST_DIR)
       set(ROOT_EVENT_DIR ${ROOTTEST_DIR}/root/treeformula/event/)
+      set(ROOT_DIR ${ROOTSYS})
```
Axel-Naumann, line 3: I don't see any remaining use of `ROOT_DIR` - can we remove that variable?

root/tree/cloning/CMakeLists.txt
```diff
@@ -1,7 +1,11 @@
...
+      execute_process(COMMAND root-config --tutdir OUTPUT_VARIABLE ROOT_TUTDIR OUTPUT_STRIP_TRAILING_WHITESPACE)
```
Axel-Naumann, line 4: This command seems identical, whether `ROOTTEST_DIR` or not. Move outside the `if()`?

guitargeek:
Superseded by:
  * https://github.com/root-project/root/pull/18795

PR #670, opened 2021-03-03
Add /build to .gitignore

eguiraud:
It is a fairly common pattern to create a build directory directly
inside small repositories. Adding the directory to .gitignore helps
tools ignore its contents.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,7 @@ AutoDict_*.cxx
 
 # /
 /root_version
+/build
 
 # /cling/cintex/
 /cling/cintex/data.root


Axel-Naumann:
What if we change files in `build/`? (And yes, we should probably rename that dir, but that should probably happen first?)

eguiraud:
There is no directory called `build` in roottest

eguiraud:
(but it would be great if we renamed the `build` directory in root :smile: )

Axel-Naumann:
Argh. Thanks. :-)

So, completely independently from this PR, do we want the same in ROOT? ;-) Apologies for the noise!

Axel-Naumann (approved):

eguiraud:
> So, completely independently from this PR, do we want the same in ROOT?

I certainly would :smile: I _think_ it would be useful in general, but it might just be my bubble

pcanal:
>inside small repositories. 

First time the roottest repository is even in the same paragraph as "small repositories" :)

PR #675, opened 2021-03-09
Build option for distributed RDF tests that use pyspark

vepadulano:
Sibling PR to https://github.com/root-project/root/pull/7413


vepadulano:
@phsft-bot build just on mac1015/default, mac11.0/default, ROOT-fedora29/default, ROOT-fedora32/default, ROOT-ubuntu18.04/default, ROOT-ubuntu2004/default with flags -Dtest_distrdf_pyspark=ON

PR #678, opened 2021-03-16
Add missing testRVecIO dependencies

chrisburr:
I found I needed to make this change to make the conda jenkins job build successfully.


eguiraud (commented):
Argh, thanks! If `ROOT::RIO` and `ROOT::TreePlayer` work that should be the "proper" way to bring in the transitive dependencies, I think. Would it be possible to give that a try?

PR #685, opened 2021-04-07
[DF] Fix test_nested_rvec_snapshot on Windows

eguiraud:
On Windows it is not allowed to delete (or, in this case, "RECREATE")
a file while it is still open. We now enclose each Snapshot call in
its own scope so the file is closed before being recreated.


bellenot (approved):
Thanks @eguiraud !

PR #686, opened 2021-04-07
Add test for ROOT_LIBRARY_PATH

pcanal:


pcanal:
@ChristianTackeGSI

pcanal:
@ChristianTackeGSI Could you take a look at the results?

ChristianTackeGSI:
Hi @pcanal !

Thanks for taking care of this!

Obviously, my test should have had more output, so that one knows, where it fails. Hmpf.

But after starring a while at it, I have some vague idea:
`/home/sftnight/build/workspace/roottest-pullrequests-build/build/lib` is listed more than once in there, which already is suspicous.
So I wonder, whether `LD_LIBRARY_PATH` (or alike) has this path in it when the tests start?

Axel-Naumann (commented):
I'd prefer a test using CMake's `env` command, which allows us to test this both on Windows and Linux/MacOS. Even if it's at the cost of having multiple subdirectories with their dedicated `.rootrc`.

Using a shall script means we will have to duplicate the whole setup for Windows, and makes debugging / reproducing / understanding the test more involved.

ChristianTackeGSI:
@Axel-Naumann

First excuse me for not taking more care of this. It was always on my TODO-list, but not at the top.

So what you're suggesting is to rewrite the shell script into a `.cmake` script that could be executed like `ROOTTEST_ADD_TEST(… COMMAND cmake ... -P ${CMAKE_CURRENT_SOURCE_DIR}/test_dynpath_setup.cmake -DROOT_root_CMD=${ROOT_root_CMD}`?
And in the `.cmake` we use `$ENV{…}`?

Interesting approach! This might work, yes!
Would take me a bit of time to rewrite. But yes, I could look into this.

pcanal:
This is moved to https://github.com/root-project/root/pull/23123

PR #691, opened 2021-04-12
Fix building libEvent.so etc on Apple M1 (v6-24)

Axel-Naumann:


pcanal (approved):

Axel-Naumann:
Works on some platforms, need to converge for 6.24, nightlies will confirm: merging

PR #694, opened 2021-04-18
Adjust `rootls -l` tests to consider key cycle

vepadulano:
Sibling PR to https://github.com/root-project/root/pull/7878 .


vepadulano:
Merging as sibling PR in ROOT has been merged

PR #695, opened 2021-04-19
Test serialization of the ROOT facade

vepadulano:
Sibling PR to https://github.com/root-project/root/pull/7886

### DIFF
--- a/python/pickle/PyROOT_readtests.py
+++ b/python/pickle/PyROOT_readtests.py
@@ -165,6 +165,21 @@ class PickleReadingComplicationsTestCase( MyTestCase ):
       f1 = TBufferFile( TBuffer.kWrite )
       f2 = pickle.loads( pickle.dumps( f1 ) )
 
+   def test3PickleFacadeCheck(self):
+      """Test serialization of the ROOT Python module.
+
+      This needs a custom __reduce__ method defined in the ROOTFacade class.
+      """
+
+      def get_root_facade():
+         return ROOT
+
+      facade = pickle.loads(pickle.dumps(get_root_facade()))
+
+      # Check attributes of the unserialized facade
+      self.assertEqual(facade.__name__, ROOT.__name__)
+      self.assertEqual(facade.__file__, ROOT.__file__)
+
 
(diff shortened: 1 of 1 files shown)


python/pickle/PyROOT_readtests.py
```diff
@@ -165,6 +165,21 @@ def test2TBufferCheck( self ):
       f1 = TBufferFile( TBuffer.kWrite )
       f2 = pickle.loads( pickle.dumps( f1 ) )
 
+   def test_return_facade(self):
```
etejedor, line 168: Can you follow the naming scheme of the other two tests (e.g. test3ReturnFacade)?

vepadulano: Right I missed that sorry! I'll wait to see the results from the CI then I'll change the name :+1:

vepadulano:
The debian10 node seems to not have picked up the correct file
```
sftnight@pcepsft11: cat /home/sftnight/build/workspace/roottest-pullrequests-build/roottest/python/pickle/PyROOT_readtests.py
...
class PickleReadingComplicationsTestCase( MyTestCase ):

   def test1RefCountCheck( self ):
      """Test reference counting of pickled object"""

      self.assertEqual( SomeCountedClass.s_counter, 0 )
      c1 = SomeCountedClass();
      self.assertEqual( SomeCountedClass.s_counter, 1 )

      c2 = pickle.loads( pickle.dumps( c1 ) )
      self.assertEqual( SomeCountedClass.s_counter, 2 )
      del c2, c1
      self.assertEqual( SomeCountedClass.s_counter, 0 )

   def test2TBufferCheck( self ):
      """Test that a TBufferFile can be pickled"""

    # the following does not assert anything, but if there is a failure, the
    # ROOT I/O layer will print an error message
      f1 = TBufferFile( TBuffer.kWrite )
      f2 = pickle.loads( pickle.dumps( f1 ) )


## actual test run
if __name__ == '__main__':
...
```
I'm updating the commit and triggering another CI job to be sure it was just a hiccup

PR #696, opened 2021-04-20
Do not overwrite ROOTSYS if set from "outside":

Axel-Naumann:
This helps e.g. with https://github.com/root-project/root/issues/7081
as *if* roottest is configures with -DROOTSYS, that will be honored
instead of taking whatever the env var $ROOTSYS points to.


Axel-Naumann:
Windows failure seems unrelated.

PR #698, opened 2021-04-23
Fix building roottest against already installed ROOT

amadio:
The variable `PYTHON_EXECUTABLE_Development_Main` is only available when roottest is being built as a subdirectory within a ROOT build. Therefore, to let roottest work again in an independent build, we need to use `PYTHON_EXECUTABLE`, and rely on ROOT's exported version of Python to just find and use the same Python version in roottest.

Ideally, we should add in ROOT a call to `find_dependency` to match this call to `find_package(Python ...)` here, so that even this is not necessary in the future.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,10 @@ endif()
 
 find_package(ROOT REQUIRED CONFIG)
 
+if(NOT DEFINED PYTHON_EXECUTABLE)
+  find_package(PythonInterp ${ROOT_PYTHON_VERSION} REQUIRED QUIET)
+endif()
+
 if(MSVC)
 
   set(libsuffix .dll)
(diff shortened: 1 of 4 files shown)


amadio:
`find_package(Python)` is only available in CMake 3.12 and above. Since roottest only needs to know which interpreter should be used, we can use the older `find_package(PythonInterp)` which should work everywhere (added and rebased).

amadio:
Trying to call `find_package(PythonInterp ...)` only when `${PYTHON_EXECUTABLE}` is not defined. Looks like calling it after calls to `find_package(Python{2,3} ...)` with CMake 3.18 causes problems.

eguiraud (approved):
LGTM, thanks @amadio . @etejedor should probably also take a look before we merge this, and the commit message should refer to https://sft.its.cern.ch/jira/browse/ROOT-10905

amadio:
Ah, indeed, I forgot to add a reference to that.

amadio:
Reference added.

CMakeLists.txt
```diff
@@ -22,6 +22,10 @@ endif()
 
 find_package(ROOT REQUIRED CONFIG)
 
+if(NOT DEFINED PYTHON_EXECUTABLE)
+  find_package(PythonInterp ${ROOT_PYTHON_VERSION} REQUIRED QUIET)
```
etejedor, line 26: I see that we do `set(ROOT_PYTHON_VERSION @PYTHON_VERSION_STRING_Development_Main@)` in `ROOTConfig.cmake.in`, so this is fine. As long as this tries to find the main/highest Python version that PyROOT has been built with, it's ok.

etejedor (approved):

PR #703, opened 2021-05-10
On Apple M1, TTreeIndex cannot handle values that large:

Axel-Naumann:
As on M1, long double == double, doing the tree formula math in double
will not have sufficient precision to keep the least significant bits
around, causing the minor value not to be found back.

Instead, reduce the number such that a M1 `long double` can still represent
it. There is a new warning in TTreeIndex() that should be triggered by
providing the original value; the test should be extended to check for
said warning to be emitted on M1 in this case.

### DIFF
--- a/root/tree/index/runindex64.C
+++ b/root/tree/index/runindex64.C
@@ -5,6 +5,16 @@
 bool test(TTree*);
 
 const char* fname = "index64.root";
+#if __APPLE__ and __arm64__
+  // Apple M1 has long double == double; these values exceed its range
+  // and cannot be represented as (even temporary) expression results.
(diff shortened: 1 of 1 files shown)


pcanal (approved):

Axel-Naumann:
@phsft-bot build just on mac11arm/default

Axel-Naumann:
@phsft-bot build just on mac11arm/default with flags -Dtmva-pymva=Off

Axel-Naumann:
@phsft-bot build

Axel-Naumann:
@phsft-bot build just on mac11arm/default

PR #705, opened 2021-05-12
Catch captured exceptions by TBB

hahnjo:
If the `std::runtime_error` is thrown by a worker thread, TBB throws a `tbb::captured_exception` on the root thread.


eguiraud:
We expect TBB to have "exact exception propagation" support, see https://sft.its.cern.ch/jira/browse/ROOT-10636

PR #706, opened 2021-05-17
On Apple M1, TTreeIndex cannot handle values that large:

Axel-Naumann:
As on M1, long double == double, doing the tree formula math in double
will not have sufficient precision to keep the least significant bits
around, causing the minor value not to be found back.

Instead, reduce the number such that a M1 `long double` can still represent
it. There is a new warning in TTreeIndex() that should be triggered by
providing the original value; the test should be extended to check for
said warning to be emitted on M1 in this case.

This is master's https://github.com/root-project/roottest/pull/703

### DIFF
--- a/root/tree/index/runindex64.C
+++ b/root/tree/index/runindex64.C
@@ -5,12 +5,12 @@
 bool test(TTree*);
 
 const char* fname = "index64.root";
-  // Apple M1 has long double == double; these values exceed its range
-  // and cannot be represented as (even temporary) expression results.
-  // There would be a warning if you'd try.
-  static constexpr bool shortlongdouble = sizeof(longdouble) < 16; // was true for __APPLE__ and __arm64__
-  const Long64_t bigval   = shortlongdouble ? 0xFFFFFFFFFFFF : 0xFFFFFFFFFFFFFFF; // still positive number
-  const ULong64_t biguval = shortlongdouble ?  0xFFFFFFFFFFFF0 : 0xFFFFFFFFFFFFFFF0; // "negative" number
+// Apple M1 has long double == double; these values exceed its range
+// and cannot be represented as (even temporary) expression results.
+// There would be a warning if you'd try.
+static constexpr bool shortlongdouble = sizeof(long double) < 16; // was true for __APPLE__ and __arm64__
+const Long64_t bigval   = shortlongdouble ? 0xFFFFFFFFFFFF : 0xFFFFFFFFFFFFFFF; // still positive number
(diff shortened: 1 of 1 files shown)


Axel-Naumann:
@phsft-bot build

pcanal:
@phsft-bot build

Axel-Naumann:
@phsft-bot build

root/tree/index/runindex64.C
```diff
@@ -5,6 +5,16 @@
 bool test(TTree*);
 
 const char* fname = "index64.root";
+#if __APPLE__ and __arm64__
```
pcanal, line 8:
```suggestion
static constexpr bool shortlongdouble = sizeof(longdouble) < 16; // was true for __APPLE__ and __arm64__
```

root/tree/index/runindex64.C
```diff
@@ -5,6 +5,16 @@
...
+  // Apple M1 has long double == double; these values exceed its range
+  // and cannot be represented as (even temporary) expression results.
+  // There would be a warning if you'd try.
+  const Long64_t bigval   = 0xFFFFFFFFFFFF;
+  const ULong64_t biguval = 0xFFFFFFFFFFFF0;
+#else
+  const Long64_t bigval   = 0xFFFFFFFFFFFFFFF;  // still positive number
+  const ULong64_t biguval = 0xFFFFFFFFFFFFFFF0;  // "negative" number
+#endif
```
pcanal, line 17:
```suggestion
  // Apple M1 has long double == double; these values exceed its range
  // and cannot be represented as (even temporary) expression results.
  // There would be a warning if you'd try.
  static constexpr bool shortlongdouble = sizeof(longdouble) < 16; // was true for __APPLE__ and __arm64__
  const Long64_t bigval   = shortlongdouble ? 0xFFFFFFFFFFFF : 0xFFFFFFFFFFFFFFF; // still positive number
  const ULong64_t biguval = shortlongdouble ?  0xFFFFFFFFFFFF0 : 0xFFFFFFFFFFFFFFF0; // "negative" number
```

pcanal (approved):

root/tree/index/runindex64.C
```diff
@@ -5,6 +5,12 @@
 bool test(TTree*);
 
 const char* fname = "index64.root";
+  // Apple M1 has long double == double; these values exceed its range
```
pcanal, line 8: On the github output, the alignment seems off (2 spaces instread of zero).

Axel-Naumann: Thanks fixed!

pcanal (approved):

PR #707, opened 2021-05-21
[DF] Modify test_snapshot so it also tests #8226

eguiraud:


eguiraud:
Rebased on top of upstream master to fix those build errors

eguiraud:
@phsft-bot build

PR #709, opened 2021-06-02
Treeformula event make to cmake

SpectreVert:
Currently switching this Makefile to Cmake. This is half of the complete tests, still need to add DrawTest.sh.


SpectreVert:
Seems like merge did something wrong.

PR #710, opened 2021-06-08
[DF] Test Snapshot of members of TBranchElement from many input files

eguiraud:
This is a test for https://github.com/root-project/root/issues/8295 (it should fail at the moment).

### DIFF
--- a/root/dataframe/CMakeLists.txt
+++ b/root/dataframe/CMakeLists.txt
@@ -224,3 +224,4 @@ ROOTTEST_ADD_TEST(test_progressiveCSV
 
 # need the '+' to autogenerate required dictionaries with ACLiC
 ROOTTEST_ADD_TEST(test_nested_rvec_snapshot MACRO test_nested_rvec_snapshot.C+)
+ROOTTEST_ADD_TEST(test_snapshot_copyaddresses MACRO test_snapshot_copyaddresses.C+)
--- a/root/dataframe/test_snapshot_copyaddresses.C
+++ b/root/dataframe/test_snapshot_copyaddresses.C
@@ -1,0 +1,63 @@
+// Regression test for issue https://github.com/root-project/root/issues/8295
+// Make sure that Snapshotting goes well when reading/writing data members of types saved as TBranchElements from
+// multiple input files (in particular, due to #8295, from the second file onwards Snapshot used to write wrong data).
+
+#include <ROOT/RDataFrame.hxx>
+#include <TError.h>
+#include <TFile.h>
+#include <TSystem.h>
+#include <TTree.h>
+
+constexpr static const char *fnamePrefix = "test_snapshot_copyaddresses";
+constexpr static int nfiles = 3;
+
+struct TwoInts {
+   int x;
+   int y;
+};
+
+void write_inputs()
+{
+   TwoInts ti{1, 2};
+   for (int i = 0; i < nfiles; ++i, ti.x += 2, ti.y += 2) {
+      const auto fname = fnamePrefix + std::to_string(i) + ".root";
+      if (gSystem->AccessPathName(fname.c_str()) == 0) // 0 means file already exists
+         continue;
+      TFile f(fname.c_str(), "recreate");
+      TTree t("t", "t");
+      t.Branch("ti", &ti);
+      t.Fill();
+      t.Write();
+   }
+}
+
+void test_snapshot_copyaddresses()
+{
+   write_inputs();
+
+   const std::string outSuffix = "_out.root";
+
+   ROOT::RDataFrame df("t", fnamePrefix + std::string("*.root"));
+   auto out_df = df.Snapshot<int, int>("t", fnamePrefix + outSuffix, {"x", "y"});
+
+   int expected = 1;
+   out_df->Foreach(
+      [&](int x, int y) mutable {
+         R__ASSERT(x == expected++);
+         R__ASSERT(y == expected++);
+      },
+      {"x", "y"});
+
+   // clean up files
+   for (int i = 0; i < nfiles; ++i) {
+      gSystem->Unlink((fnamePrefix + std::to_string(i) + ".root").c_str());
+   }
+   gSystem->Unlink((std::string("_out") + fnamePrefix + ".root").c_str());
+}
+
+#ifndef __CLING__
+int main()
+{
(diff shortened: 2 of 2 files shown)


root/dataframe/test_snapshot_copyaddresses.C
```diff
@@ -0,0 +1,63 @@
...
+   const std::string outSuffix = "_out.root";
+
+   ROOT::RDataFrame df("t", fnamePrefix + std::string("*.root"));
+   auto out_df = df.Snapshot<int, int>("t", fnamePrefix + outSuffix, {"x", "y"});
+
+   int expected = 1;
+   out_df->Foreach(
+      [&](int x, int y) mutable {
+         R__ASSERT(x == expected++);
+         R__ASSERT(y == expected++);
```
Axel-Naumann, line 47: Isn't `R__ASSERT` only functional for debug builds? Consider using `exit(1)` with a prior `std::cerr <<`

root/dataframe/test_snapshot_copyaddresses.C
```diff
@@ -0,0 +1,63 @@
...
+      },
+      {"x", "y"});
+
+   // clean up files
+   for (int i = 0; i < nfiles; ++i) {
+      gSystem->Unlink((fnamePrefix + std::to_string(i) + ".root").c_str());
```
Axel-Naumann, line 53: If we fail above then we leave the output file. Will that be a problem for the next run?

root/dataframe/test_snapshot_copyaddresses.C
```diff
@@ -0,0 +1,63 @@
...
+{
+   write_inputs();
+
+   const std::string outSuffix = "_out.root";
+
+   ROOT::RDataFrame df("t", fnamePrefix + std::string("*.root"));
+   auto out_df = df.Snapshot<int, int>("t", fnamePrefix + outSuffix, {"x", "y"});
+
+   int expected = 1;
+   out_df->Foreach(
```
Axel-Naumann, line 44: Shouldn't you be (also) testing the content of the `Snapshot`ed output file?

root/dataframe/test_snapshot_copyaddresses.C
```diff
@@ -0,0 +1,63 @@
...
+   for (int i = 0; i < nfiles; ++i) {
+      gSystem->Unlink((fnamePrefix + std::to_string(i) + ".root").c_str());
+   }
+   gSystem->Unlink((std::string("_out") + fnamePrefix + ".root").c_str());
+}
+
+#ifndef __CLING__
+int main()
+{
+   test_snapshot_copyaddresses();
```
Axel-Naumann, line 61:
```suggestion
   return test_snapshot_copyaddresses();
```
would be nice, and make `test_snapshot_copyaddresses()` return 0 on success,

root/dataframe/CMakeLists.txt
```diff
@@ -224,3 +224,4 @@ ROOTTEST_ADD_TEST(test_progressiveCSV
 
 # need the '+' to autogenerate required dictionaries with ACLiC
 ROOTTEST_ADD_TEST(test_nested_rvec_snapshot MACRO test_nested_rvec_snapshot.C+)
+ROOTTEST_ADD_TEST(test_snapshot_copyaddresses MACRO test_snapshot_copyaddresses.C+)
```
Axel-Naumann, line 227: Why do you ACLiC this, is it needed?

Axel-Naumann (approved):
Just a couple of comments!

root/dataframe/test_snapshot_copyaddresses.C
eguiraud: yes, see my question at https://mattermost.web.cern.ch/root/pl/rku4nod8dp8t7ekb7wtmhf3jqw , it would be nice to have an equivalent of gtest's `EXPECT_TRUE`. I will change this to log + exit

root/dataframe/test_snapshot_copyaddresses.C
eguiraud: potentially yes, in case the output file is malformed. however the non-RAII behavior here is intentional: it is useful to have the files used by a failed test there, to check what's happened.

Given these two things, I think I should remove the check above that only creates the output files if they don't already exist, and always overwrite them, what do you think?

root/dataframe/test_snapshot_copyaddresses.C
eguiraud: this is testing the contents of the snapshotted output file

root/dataframe/test_snapshot_copyaddresses.C
eguiraud: that only makes sense if I make it return non-zero otherwise, but we are considering the `exit(1)` behavior instead?

root/dataframe/CMakeLists.txt
eguiraud: it's just a copy-paste from the line above. apparently at some point Danilo or I thought we needed to ACLiC the test to be able to perform I/O of custom types defined in the test source (see the comment two lines above), but that's not really the case isn't it?

root/dataframe/test_snapshot_copyaddresses.C
eguiraud: now we always overwrite the input files

root/dataframe/CMakeLists.txt
pcanal: Yes, it is needed for that. (i.e. we technically do not (fully) support I/O of interpreted classes.

Axel-Naumann: D'oh of course, thanks!

root/dataframe/test_snapshot_copyaddresses.C
Axel-Naumann: `root test_snapshot_copyaddresses.C+` will exit 0 if `test_snapshot_copyaddresses()` returns 0. If `test_snapshot_copyaddresses` returns `void` then I *think* it will exit 0, so yes, you're probably right :-)

eguiraud:
Closing in favor of #717 which contains the exact same set of changes (including those that address the review comments) but uses a branch name in sync with master and has already squashed things together.

PR #711, opened 2021-06-09
[build] clean hist Makefile

SpectreVert:


SpectreVert:
Verified that no tests are lost during merge.

PR #714, opened 2021-06-09
[build] clean collection Makefile

SpectreVert:


SpectreVert:
Verified that no tests are lost during merge.

pcanal:
How did you verify?

SpectreVert:
I added some gibberish at the start of the Makefile, then compiled. Did not yield an error.
Also verified the number of tests didn't change after the Makefile was removed and the project compiled.

pcanal:
> Also verified the number of tests didn't change after the Makefile was removed and the project compiled.

As a side note for when you start converting Makefile into CMakeFile.  This technique works in this case since there is one test per makefile (so if the makefile was run before and is no longer run, it would decrease the number of test by one).

This is not the case when replace the MakeFile into a CMakeFile.  Each MakeFile will be replaced by (usually) more than one ctest tests (so will increase the number of test reported).  In addition many test are not explicitly listed in the Makefile (the makefile detect the presence of a specific file pattern and use them).  For each makefile being replace, you probably will need to run the makefile by hand to see the output and make a note of the list of actual test.

pcanal (approved):

SpectreVert:
Thanks for the feedback -- for this one I wasn't too worried because it looked like it had been moved to CMakeFile a while back. 

However for the ones that need translation from Makefile to CMakeFile I try to make sure to do as you specified, checking for hidden or implicit tests beforehand to not miss one.

PR #715, opened 2021-06-09
[build] clean html Makefile

SpectreVert:


SpectreVert:
Verified that no tests lost during merge.

PR #716, opened 2021-06-09
move treeformula/event Makefile over to CMake

SpectreVert:


root/treeformula/event/dt_wrap.C
```diff
@@ -1,5 +1,6 @@
+#include "dt_RunDrawTest.C"
+
 void dt_wrap(const char* from, Int_t mode = 0, Int_t verboseLevel = 0) {
-   gROOT->ProcessLine(".L dt_RunDrawTest.C+");
```
Axel-Naumann, line 2: That's a change of functionality - if possible please leave the original version, and add the missing `#include`s to `dt_RunDrawTest.C`

Axel-Naumann: @pcanal remind me - why do we need to pre-build the ACLiC library through GNUMake instead of "just" letting ACLiC do its job?

pcanal: This has 2 advantages
* If the build fails, it stops at its own stage
* Reduce the log pollution in case of any warning or compiler messages.

Axel-Naumann: > If the build fails, it stops at its own stage

Agreed (at the cost of less coverage of ACLiC). OTOH when switching to CMake, having test-prerequisites built as a separate step might be more of a pain than a gain: as you know, `ctest foo` doesn't (necessarily) build foo's dependencies...

> Reduce the log pollution in case of any warning or compiler messages.

This one I find problematic. We are not seeing many of the warnings in our test code, sometimes making failure analysis more difficult.

But let's discuss by voice - maybe tomorrow. Thanks for reminding me of the rationale for the current setup!

oshadura:
woooow! looks like you did it? Congrats!!!

oshadura:
@SpectreVert please convert it from draft?

bellenot (approved):
Fine with me if everything is green and if @Axel-Naumann is happy with it (since there was an unresolved discussion about some changes)

root/treeformula/event/CMakeLists.txt
```diff
@@ -1,7 +1,94 @@
...
+#
+# Generate tests using file generated with EVENT_GENERATE_FILE. dt_wrap.C
+# is run with differing modes (0, 1, 2, 3, 4)
+#
+function(EVENT_GENERATE_TESTS file_name)
+  foreach(mode 0;1;2;3;4)
+    ROOTTEST_ADD_TEST("${file_name}-${mode}"
+      COMMAND ${ROOT_root_CMD} -q -b -l "${CMAKE_CURRENT_SOURCE_DIR}/dt_wrap.C\(\"${file_name}.root\",${mode}\)"
+      DEPENDS EventGeneration
+      RUN_SERIAL)
```
pcanal, line 72: Is this `RUN_SERIAL` necessary?

SpectreVert: This one is tricky. I'd like to remove it, but from the tests I've done, even `DEPENDS EventGeneration "generation-${file-name}"  read write` lead to spurious fails and I couldn't get it to execute properly without `RUN_SERIAL`

oshadura (commented):
No of these failures are related to changes (and LGTM from me)

SpectreVert:
@bellenot Since you approved a few days back, everything looks good to you?

bellenot:
> @bellenot Since you approved a few days back, everything looks good to you?

As far as I can tell, yes.

PR #718, opened 2021-06-10
[build] clean math Makefile

SpectreVert:
[todo] translate mathcore/Makefile


SpectreVert:
Verified no tests are being lost during merge.

Axel-Naumann (approved):
LGTM!

And while we're at it, please convert https://github.com/root-project/root/blob/master/math/genvector/test/coordinates3D.cxx etc to proper gtest unit tests - that way, we can just remove all of roottest/root/math/genvector

PR #721, opened 2021-06-10
Fix test_snapshot_copyaddresses.C on Windows

bellenot:
Fix issue with files not being deleted on Windows due to file handles not yet released by the process

### DIFF
--- a/root/dataframe/test_snapshot_copyaddresses.C
+++ b/root/dataframe/test_snapshot_copyaddresses.C
@@ -36,26 +36,29 @@ void test_snapshot_copyaddresses()
    write_inputs();
 
    const std::string outSuffix = "_out.root";
+   // On Windows, one cannot delete files while they are still used (not released)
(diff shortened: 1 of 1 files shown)


eguiraud (changes requested):
Hi @bellenot , thank you, I'm still surprised that `Unlink` fails without the usual message that says "cannot delete already opened file" or something like that, but :man_shrugging: 

Just two small things:
- the indentation of the code inside braces is now wrong :smile:
- please add a comment mentioning why we need these seemingly no-op braces in the code, or someone might "optimize" them away in the future

bellenot:
@eguiraud OK, thanks. I'll correct the indentation and add a comment

eguiraud (approved):
Thanks a lot Bertrand!

bellenot:
@eguiraud You're very welcome! :-)

bellenot:
The failure is unrelated. Merging

PR #722, opened 2021-06-18
[PyROOT] Add test for cross-inheritance object pickling

etejedor:
Test for https://github.com/root-project/root/pull/8476


vepadulano (approved):
Nice thanks!!!!

PR #724, opened 2021-06-29
remove legacy script addtest which is not used anymore

SpectreVert:
File has not been modified since 2005 and is not used anywhere. Clutter that should be removed.


Axel-Naumann (approved):

Axel-Naumann:
Test failure unrelated

PR #725, opened 2021-06-29
remove legacy script mktest which is not used anymore

SpectreVert:
File has not been modified since 2004 and is not used in the build. Clutter that should be removed.


Axel-Naumann (approved):

Axel-Naumann:
Test failure unrelated

PR #726, opened 2021-06-29
remove legacy Makefile which is not run anywhere

SpectreVert:
File has been superseded by `Makefile`. Has been verified not to run under a build. Should be removed


pcanal:
Please remove this only as the very last step.  It is still used(/useable) to clean all the by-product of the 'make' runs that are executed by ctest.

pcanal:
It is still used manually to 'use' the makefile and thus this should be the last thing to go.

guitargeek:
Closed in favor of a corresponding PR in the main ROOT repository:
  * https://github.com/root-project/root/pull/18596

PR #734, opened 2021-07-07
treeformula/clones Makefile to CMakeFile

SpectreVert:
I've translated the first and only test I could extract from the Makefile. 

@Axel-Naumann @pcanal please tell me what other tests need to be (re)made, as the naming convention and documentation of this directory is not exactly optimal.


pcanal:
Running `make VERBOSE=true` will give the complete list of active test.

Namely there is a customized test (`nestedClones2`), a regular implicit test (`nestedClones3.C`) and there is also a disable test (`nestedClones`)

Do we have a plan to preserve the information content of the disabled test (it could be to also migrate them to `CMakeList` but disabled them or it could be to analyze them a bit to open a tickets (or add the test as information to an existing ticket) or something else).

Note that the list of implicit test rule can be extracted form `roottest/script/Rules.mk`, the most prominent are trigger by a file named:

1. exec*.cxx
2. exec*C
3. assert*cxx
4. assert*C
5. run*cxx
6. run*C
7. exec*py

Cheers,
Philippe.

Axel-Naumann:
https://github.com/root-project/root/issues/8744 is addressing the "FAIL" test. Please `git rm runNestedClones.C` as part of this PR

Axel-Naumann (changes requested):
Thanks! See https://github.com/root-project/roottest/pull/734#issuecomment-887504726 for what's left to do

pcanal:
@phsft-bot build

pcanal (approved):

pcanal:
The failure seems unrelated.

PR #736, opened 2021-07-13
Fix the CMAKE_CXX_STANDARD value on Windows

bellenot:


CMakeLists.txt
```diff
@@ -51,6 +51,14 @@ if(MSVC)
 
   include(${ROOT_USE_FILE})
   set(ROOTSYS ${ROOTSYS} CACHE INTERNAL "")
+  exec_program(root-config.bat ARGS "--features" OUTPUT_VARIABLE ROOT_FEATURES RETURN_VALUE RETVAR)
```
pcanal, line 54: Out of curiosity (and the answer may deserve to be added as a comment), why is `root-config.bat` the authoritative answer and the c++ standard selected (when calling cmake)?

bellenot: I simply copied the method from non-Win32 part (scroll down), using `root-config.bat` instead of the `root-config` shell script

pcanal (approved):

CMakeLists.txt
pcanal: right ... most importantly I missed this was for `roottest` rather than `root`..

PR #737, opened 2021-07-21
revert "move treeformula/event Makefile over to CMake (#716)"

SpectreVert:
Reverting changes awaiting implementation of https://github.com/root-project/root/issues/8707

As discussed problem looks like a dependency issue which will be fixed after ctest fixtures are implemented into the CMake macros.


SpectreVert:
@Axel-Naumann looks like the build had no trouble with the old Makefile tests. Should we merge now?

Axel-Naumann:
Thanks, Costa!

PR #741, opened 2021-07-23
remove old, unused Makefile

SpectreVert:
already has been translated to CMakeLists.txt


Axel-Naumann (approved):
`ClingWorkAroundUnnamedIncorrectFileLoc` is indeed flxed / not needed anymore. We should update array1test_convert.sh accordingly.

PR #743, opened 2021-07-23
cint has been deprecated; remove folder

SpectreVert:
@Axel-Naumann @pcanal can you confirm this folder is no longer needed?


Axel-Naumann (approved):
As `CMakeLists.txt` puts it: `if (ROOT_cintex_FOUND)`. So indeed these tests were never run since cling. The actual tests resided in `$ROOTSYS/cint/cintex/test`, a directory that doesn't exist anymore since years. LGTM, thanks!

SpectreVert:
That's what I figured. Perfect.

PR #763, opened 2021-08-06
[py-cpp] Force address to be unsigned:

Axel-Naumann:
This matches `array("L",...` and fixes a signedness mismatch on 32bit:
```
FAIL: Test passing of variants of void pointer arguments                
----------------------------------------------------------------------                                                              
Traceback (most recent call last):                                 
  File "/home/sftnight/build/AXEL/src/roottest/python/cpp/PyROOT_cpptests.py", line 62, in test12VoidPointerPassing
    self.assertEqual( addressofo.buffer_info()[0], Z.GimeAddressPtrPtr( addressofo ) )
AssertionError: 3079933240 != -1215034056   
```

FYI @wlav - I'd be interested to hear your opinion on this potential change to the code you wrote 14 years ago, if you find the time. I'm happy to hear that any time, also post-merge. This fixed a current test failure so we'll have to accelerate pre-merge review, assuming this passes everywhere.

### DIFF
--- a/python/cpp/PointerPassing.C
+++ b/python/cpp/PointerPassing.C
@@ -4,30 +4,30 @@ public:
       static Z z[2];
       return &(z[idx]);
    }
-   static long GimeAddressPtr( void* obj ) {
-      return reinterpret_cast< long >( obj );
+   static unsigned long GimeAddressPtr( void* obj ) {
+      return reinterpret_cast< unsigned long >( obj );
    }
 
-   static long GimeAddressPtrRef( void*& obj ) {
-      return reinterpret_cast< long >( obj );
+   static unsigned long GimeAddressPtrRef( void*& obj ) {
+      return reinterpret_cast< unsigned long >( obj );
    }
 
-   static long SetAddressPtrRef( void*& obj ) {
+   static unsigned long SetAddressPtrRef( void*& obj ) {
       obj = (void*)0x1234;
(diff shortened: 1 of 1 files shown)


vepadulano (approved):
LGTM, thanks for the fix :smile:

wlav:
> FYI @wlav - I'd be interested to hear your opinion on this potential change to the code you wrote 14 years ago

You can always look up my current take on these old tests in cppyy/test.

This test, however, is for a completely obsolete workaround and thus does not matter: it was a trick to create some storage containing an address when `void*` was still returned as an integer representing the pointer value. These days, however, `void*` is a `cppyy.LowLevelView` which can be manipulated and passed through a `void*&` with the expected behavior. And even if you wanted to work with integers, you'd use `ctypes.c_void_p` (which is newer than this test/workaround) instead of `array.array` b/c the latter does not have an 8-byte integer type to chose from on 64b MS Windows. (You could use `numpy.ndarray` instead, but here too, this test actually predates `numpy`.)

Axel-Naumann:
Thank you for the explanation , @wlav - that's sort of what I expected, also given that I could not find the test back in cppyy.

wlav:
> also given that I could not find the test back in cppyy.

Test is here: https://github.com/wlav/cppyy/blob/master/test/test_advancedcpp.py#L393

PR #766, opened 2021-08-10
disable `rlibmap` test when using libasan

SpectreVert:
Test is failling on ASAN build- disabling it should fix it


bellenot (commented):
IIUC this test fails with asan. Are you sure it's not a real issue?

guitargeek:
If the test fails with ASAN, this needs to be investigated and fixed instead of disabling the test. But thank you very much for highlighting the problem :pray:

PR #772, opened 2021-08-26
root/treeformula/references make to cmake

SpectreVert:
Translation made to CMake ; in relation with https://github.com/root-project/root/issues/8912


guitargeek (approved):
Thank you very much, especially for resurrecting the `execRefClonesArrayTest.cxx` test on the go!

PR #775, opened 2021-08-31
[RDF] Serialize subtests of header distribution in DistRDF

etejedor:
This prevents "headerX.hxx file not found" errors when both subtests try to concurrently clean the headers at cleanup time from different threads.

https://lcgapp-services.cern.ch/root-jenkins/job/root-nightly-master/LABEL=ROOT-ubuntu18.04,SPEC=nortcxxmod,V=master/2956/testReport/projectroot.roottest.python/distrdf/roottest_python_distrdf_spark_test_include_headers/


vepadulano (approved):
Nice catch, thanks!

PR #785, opened 2021-10-21
[DF] Test optional arguments of distributed Snapshot, AsNumpy

vepadulano:
Sibling PR of https://github.com/root-project/root/pull/9153


vepadulano:
Test failure might be related to the comparison between Python strings and std::string

PR #792, opened 2021-11-09
Fix several tests on Win64 (use Longptr_t + pointer formatting)

bellenot:


root/collection/execHistArray.cxx
```diff
@@ -43,7 +43,7 @@ void write(const char *filename = "HistArray.root")
    f->WriteObjectAny(ti, "TArrayI", "ti");
 
    TArrayL* tl = new TArrayL(10);
-   Long_t al[5] = { 1000L, 10000L, 100000L, 1000000L, 2000000L };
+   Longptr_t al[5] = { 1000L, 10000L, 100000L, 1000000L, 2000000L };
```
pcanal, line 46: This one seems odd.  Those value does not seem to be addresses and `TArrayL` should be handling `Long_t` ... what am I missing?

bellenot: you might be right, I'm checking

bellenot: So here we go:
```
1346: C:\Users\sftnight\git\roottest\root\collection\execHistArray.cxx:47:17: error: cannot initialize a parameter of type 'Longptr_t *' (aka 'long long *') with an lvalue of type 'Long_t [5]'
1346:    tl->Adopt(5, al);
1346:                 ^~
1346: C:/Users/sftnight/build/x64/preview\include\TArrayL.h:39:44: note: passing argument to parameter 'array' here
1346:    void          Adopt(Int_t n, Longptr_t *array);
1346:                                            ^
```
Should I revert the changes in `TArrayL` ?

pcanal: Yes, I don't think TArrayL/X should be used for pointers, only for numbers.

pcanal (approved):

PR #795, opened 2021-12-01
[DF] Move Spark unittests to roottest and add Dask equivalent

vepadulano:
Sibling PR of https://github.com/root-project/root/pull/9361 . Adds Spark and Dask unittests


python/distrdf/dask/test_backend.py
```diff
@@ -0,0 +1,245 @@
...
+        # Dummy df just to retrieve the initialization function
+        df = Dask.RDataFrame(10, daskclient=self.client)
+        f = df._headnode.backend.initialization
+
+        self.assertEqual(f(), 123)
+
+    def test_initialization_method(self):
+        """
+        Check initialization method in Dask backend.
+        Define a method in the ROOT interpreter called getValue which returns
```
etejedor, line 129: I think this is rather defining an integer `userValue` with a given value?

python/distrdf/dask/test_backend.py
```diff
@@ -0,0 +1,245 @@
...
+        cls.client = Client(LocalCluster(n_workers=2, threads_per_worker=1, processes=True))
+
+    @classmethod
+    def tearDownClass(cls):
+        """Reset test environment."""
+        cls.client.shutdown()
+
+    def test_change_attribute_when_npartitions_greater_than_clusters(self):
+        """
+        Check that the `npartitions class attribute is changed when it is
```
etejedor, line 213: Missing closing backtick

python/distrdf/dask/test_backend.py
```diff
@@ -0,0 +1,245 @@
...
+        nentries = histo.GetEntries()
+
+        self.assertEqual(nentries, 100)
+        self.assertEqual(df._headnode.npartitions, 1)
+
+        os.remove(filename)
+
+    def test_user_supplied_npartitions_have_precedence(self):
+        """
+        Check that even if Dask configuration options could optimize the number
```
etejedor, line 234: By `Dask configuration options` here you mean the internal optimization algorithm of the Dask backend for the number of partitions?

vepadulano: Mmm yeah sort of. In this case the dask client is connected to a `LocalCluster` with 2 cores, but the user passes `npartitions=4` so it should still stay at 4 even though the backend's `optimize_npartitions` would return 2

vepadulano: I'll improve the docs

vepadulano:
The timeout on the mac11 node is very weird. It is not reproducible on any other configuration. I suspect it might have to do with creating a Dask `LocalCluster` with all the cores of the machine. I will add a commit to test this

PR #802, opened 2022-01-11
Fix broken array element - thanks, clang!

Axel-Naumann:


hahnjo (approved):
Ha, I was also looking into that. Consider me impressed by Clang's warning capabilities (though not sure why this only shows up on macOS 12)

pcanal (approved):

PR #805, opened 2022-01-21
[DF] Add test for Redefine in distributed RDataFrame

vepadulano:
Sibling PR to https://github.com/root-project/root/pull/9658

### DIFF
--- a/python/distrdf/spark/test_reducer_merge.py
+++ b/python/distrdf/spark/test_reducer_merge.py
@@ -302,6 +302,21 @@ class ReducerMergeTest(unittest.TestCase):
         for filename in tmp_files:
             os.remove(filename)
 
+    def test_redefine_one_column(self):
+        """Test that values of one column can be properly redefined."""
+        # A simple dataframe with ten sequential numbers from 0 to 9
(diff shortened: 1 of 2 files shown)


python/distrdf/spark/test_reducer_merge.py
```diff
@@ -302,6 +302,21 @@ def test_distributed_snapshot_columnlist(self):
         for filename in tmp_files:
             os.remove(filename)
 
+    def test_redefine_one_column(self):
+        """Test that values of one column can be properly redefined."""
+        # A simple dataframe with ten sequential numbers from 0 to 9
+        df = Spark.RDataFrame(10, sparkcontext=self.sc)
+        df_before = df.Define("x", "1")
```
etejedor, line 309: Might be good to check if we can have tests that are generic and can run for all backends, like this we would avoid the repetition (in this case Spark and Dask).

vepadulano: The duplication is there to test that actually both Spark and Dask can run the same test. I think making generic tests could be done, we would need a common, mock backend only for such tests. We can discuss if this is needed and draft another PR

PR #807, opened 2022-01-25
[win] Remove forced includes w32pragma.h and sehmap.h

bellenot:
Following the recent changes in ROOT (commit #46f3742), the standalone roottest fails when building googletest:
```
roottest\googletest-prefix\src\googletest-build\CMakeFiles\CMakeTmp\testCCompiler.c : fatal error C1083: Cannot open include file: 'w32pragma.h': No such file or directory
```
This commit fixes the error. `w32pragma.h` and `sehmap.h` are not needed anyway


pcanal (approved):

bellenot:
The reported windows error is from the commit log, and not from the output. Merging.

PR #809, opened 2022-01-25
Try to fix #9600 Missing dependency when building roottest as part of…

bellenot:
… ROOT


hahnjo:
Unfortunately this doesn't work for out-of-tree builds as predicted by @pcanal. I'm in the last stage of testing a different fix.

bellenot:
> Unfortunately this doesn't work for out-of-tree builds as predicted by @pcanal. I'm in the last stage of testing a different fix.

OK, thanks

PR #810, opened 2022-01-25
stl_makeproject_test: Add dependency on Hist (if available)

hahnjo:
This resolves issues with in-tree builds of roottest where the build system would try to generate the dictionary before Hist is built and then fails because `TH1D.h` is included, but `Hist.pcm` not specified as a by-product.

Closes https://github.com/root-project/root/issues/9600


bellenot (commented):
Fine with me if @pcanal agrees. But how does it work outside then? How is the dependency on Hist resolved? Because I suppose the header still includes TH1D?

hahnjo:
> But how does it work outside then? How is the dependency on Hist resolved? Because I suppose the header still includes TH1D?

In case of an out-of-tree build, we can rely that ROOT has been fully built, so we don't need a dependency.

bellenot:
Thanks @hahnjo !

PR #818, opened 2022-02-04
[DF] Actually run roottest distrdf tests of distributed RunGraphs

vepadulano:
In roottest, python tests need to include a call to actually start the
test, unlike plain pytest tests that would just run without it.

When introducing distributed RunGraphs tests, this call was omitted,
effectively preventing CI tests of the feature.


vepadulano:
Test failures are unrelated and fixed in latest commit on master

PR #819, opened 2022-02-04
Minor fix to `meta/tclass/execState.C`

jalopezg-r00t:
Use `gInterpreter->Declare()`, instead of `gROOT->ProcessLine()`, to introduce a forward declaration.  By design, `ProcessLine()` cannot guarantee that the user-provided code is not internally modified by cling.

See https://github.com/root-project/root/pull/9782.

### DIFF
--- a/root/meta/tclass/execState.C
+++ b/root/meta/tclass/execState.C
@@ -6,7 +6,7 @@ void writeState() {
 int execState() {
    int result = 0;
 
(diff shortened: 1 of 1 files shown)


pcanal (approved):
I agree this is a better version of the test but it looks like (see https://github.com/root-project/root/pull/9782#issuecomment-1030056862) there is a real issue here.

PR #820, opened 2022-02-08
Link libEvent against dependent libs:

Axel-Naumann:
in the old Makefile, many of the old variables are unset when running
from ctest. One of those is EXPLLINKLIBS. Instead, the dependent libraries
are linked through $^ - but only on Linux, not on macOS.

Fix roottest/root/treeformula/sync failing in Simple with
```
cling::DynamicLibraryManager::loadLibrary(): dlopen(/Users/sftnight/build/jenkins/night/LABEL/mac12/SPEC/soversion/V/master/roottest/root/treeformula/sync/libEvent.so, 0x0009): symbol not found in flat namespace '__ZN4TH1FC1EPKcS1_idd'
Processing Run.C(1)...
...
```
due to the missing dependent libraries (libHist).

That's fixed simply by using the default case for both Linux and macOS. Yay!

### DIFF
--- a/scripts/Event.mk
+++ b/scripts/Event.mk
@@ -48,9 +48,6 @@ else
 ifeq ($(ARCH),aix5)
 		$(CMDECHO) /usr/vacpp/bin/makeC++SharedLib $(OutPutOpt) $@ $(LIBS) -p 0 $^
 else
-ifeq ($(PLATFORM),macosx)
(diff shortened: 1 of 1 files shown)


Axel-Naumann:
This needs to go into 6.26, too.

pcanal (approved):

pcanal:
Both mac build ran successfully to completion.  Seemingly waiting on Windows node.

PR #822, opened 2022-02-10
cleantest: depend on clean; run test as target:

Axel-Naumann:
The approach before evaluates CLEAN_TARGETS even though many
of the "calling" Makefiles only set it after the inclusion of
Rules.mk.


pcanal (approved):
See commit fc360ac615692e791c5706c620cdb9e21936e23d.  ince the depth of each make cleantest is now much shallower than it used to be (and we don't use cygwin anymore) and the makfiles are on their way out, I think this is fine.

Axel-Naumann:
Failures seem unrelated

PR #835, opened 2022-02-25
Fix ASan complaint in root/addresses/iobug.C

hahnjo:
Commit 6418a9d9a7 copied parts of `TGraph` and `TGraphErrors`, but not the code to resize the member arrays of `GraphErrors` when setting points beyond the initially specified size. Assume that this isn't relevant for the test.

Fixes https://github.com/root-project/root/issues/8917


hahnjo:
ping @pcanal

pcanal:
This seems to disable per se part of the test.   It also seems that there was some attempts at bringing-in the array resizing (`Graph::AllocateArrays`) so it seems that this might be just 'hiding' the 'real' problem (with the test).

hahnjo:
Yes, this is what I wrote in the commit message:
> Assume that this isn't relevant for the test.

If that's not the case, I can close this PR and somebody needs to properly fix this. To reiterate, the problem is that `GraphErrors` doesn't know how to resize its arrays, so ROOT IO reads out of bounds because `fNpoints` in the base class is increased.

hahnjo:
@pcanal what do we do here? Is this a valid "solution" or shall I close the PR?

pcanal:
@couet Could you take a closer look a this problem?  eg.  I am not sure what the ASAN report actually is and whether it is because Graph::AllocateArrays is not working or something else.

pcanal:
@couet nevermind :), after a second look, I think I know where to take this (the test needs to be 'fixed' to properly extend the arrays as needed).

pcanal:
The missing memory allocation are added by https://github.com/root-project/roottest/pull/897

PR #838, opened 2022-03-01
[DF] Run distributed tests in sequence

vepadulano:
CTest by default runs tests in parallel. Distributed tests need more
resources by default, plus they all need to create instances of mock
local clusters that may need to reserve ports. On some nodes this leads
to timeouts when running in parallel.

### DIFF
--- a/python/distrdf/common/CMakeLists.txt
+++ b/python/distrdf/common/CMakeLists.txt
@@ -20,6 +20,7 @@ if (ROOT_test_distrdf_pyspark_FOUND AND ROOT_test_distrdf_dask_FOUND)
 
     ROOTTEST_ADD_TEST(common_test_rungraphs
                     MACRO test_rungraphs.py
-                    ENVIRONMENT ${PYSPARK_ENV_VARS})
+                    ENVIRONMENT ${PYSPARK_ENV_VARS}
+                    RUN_SERIAL)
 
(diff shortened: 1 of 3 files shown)


python/distrdf/common/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ if (ROOT_test_distrdf_pyspark_FOUND AND ROOT_test_distrdf_dask_FOUND)
 
     ROOTTEST_ADD_TEST(common_test_rungraphs
                     MACRO test_rungraphs.py
-                    ENVIRONMENT ${PYSPARK_ENV_VARS})
+                    ENVIRONMENT ${PYSPARK_ENV_VARS}
```
etejedor, line 23: This is quite a few tests to run in serial mode. The problem is that they would run alone with respect to all other tests, not only the distributed RDF tests.

Could we unite all the Dask ones and all the Spark ones under one common test module each, which create one Dask client and one Spark scheduler, respectively, and run all their tests with that? Ideally we would still keep the separate modules for different tests, but have one common supermodule that imports them?

vepadulano: Thanks for the hint, I will try that

vepadulano:
Closing in favor of #847

PR #845, opened 2022-03-10
Fix perfstattest-libevent-build on macOS

hahnjo:
Remove the flag to always rebuild with Unix Makefiles to match what we do in `cling-stl-dicts`, which also has a library depending on a generated dictionary file (actually multiple). The reason this does not work is that `/fast` ignores dependencies, and in conjunction with `--always-make` it results in the build system not finding the generated files.


Axel-Naumann:
THANK YOU! I have already spent too many hours on this one, super cool that you found the cause!

hahnjo:
Well, there's an obvious downside: If we now run the test repeatedly, it won't actually do anything. But I *think* that happens with Ninja anyway, so no reason to let the tests fail with Unix Makefiles.

bellenot (approved):
LGTM, and I trust you 🙂

PR #854, opened 2022-03-29
Remove the RUN_SERIAL flag for MSVC

bellenot:
After the [commit #fcec2fc3](https://github.com/root-project/root/commit/fcec2fc38fec87352469c0550aa68993b709172e) in ROOT, this flag is not necessary anymore on Windows. This should speed-up roottest significantly


bellenot:
Error is unrelated. Merging

PR #871, opened 2022-04-28
Remove an old test for TF1

lmoneta:
The test is not adding anything new to already existing tests of TF1,
for example in stressHistoFit.


dpiparo (approved):
LGTM. Thanks for this PR. Proactively removing tests that do not spark joy any more will translate in substantial saving of resources in the medium term!

PR #876, opened 2022-05-16
More fixes for parallel builds on Windows

bellenot:
Using the `ROOT::` prefix for libraries make the build system check if the libraries have to be rebuilt and introduce extra time and potential errors when building in parallel. For example, `roottest-root-meta-loadernotapp-build` takes 72.24 sec with `LIBRARIES ROOT::Core ROOT::Hist` and only 15.87 sec with `LIBRARIES Core Hist` (and it's even worse on Win64)
Without the `ROOT::` prefix (`LIBRARIES Core Hist`):
```
C:\Users\sftnight\build\release>ctest -VV -C Release -R roottest-root-meta-loadernotapp-build
UpdateCTestConfiguration  from :C:/Users/sftnight/build/release/DartConfiguration.tcl
Parse Config file:C:/Users/sftnight/build/release/DartConfiguration.tcl
 Add coverage exclude regular expressions.
UpdateCTestConfiguration  from :C:/Users/sftnight/build/release/DartConfiguration.tcl
Parse Config file:C:/Users/sftnight/build/release/DartConfiguration.tcl
Test project C:/Users/sftnight/build/release
Constructing a list of tests
Ignore test: test-tcollex
Ignore test: roottest-cling-parsing-semicolon
Ignore test: roottest-root-rint-TabCom
Ignore test: roottest-root-rint-BackslashNewline
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 1783
    Start 1783: roottest-root-meta-loadernotapp-build

1783: Test command: "C:\Program Files\CMake\bin\cmake.exe" "--build" "C:/Users/sftnight/build/release" "--config" "Release" "--target" "loadernotapp" "--"
1783: Environment variables:
1783:  ROOTSYS=C:/Users/sftnight/build/release
1783:  PYTHONPATH=C:/Users/sftnight/build/release/bin
1783: Test timeout computed to be: 1500
1783: Microsoft (R) Build Engine version 16.11.2+f32259642 for .NET Framework
1783: Copyright (C) Microsoft Corporation. All rights reserved.
1783:
1783: Build started 16/05/2022 14:25:03.
1783: Project "C:\Users\sftnight\build\release\roottest\root\meta\loadernotapp.vcxproj" on node 1 (default targets).
[...]

PR #877, opened 2022-05-18
stdpair: Fix building the test's library on MacOS.

pcanal:
See also #858.

Note technically only one of the two commits:

a. Build libraries in the ROOT build step
b. Add explicit dependencies to correctly build the libraries when `ctest` is running

is needed to solve the problem.

Tests in `roottest` have been following those 2 distinct implementations (seemingly arbitrarily).  They require 2 different pattern of use to get things to work after changing files (related to 'libraries') in `roottest`:

1. `make && ctest`
or
2. just `ctest`

As we have it know, because several tests implement (a), we technically have to use pattern `1.` ; however if working on specific tests that implement (b) pattern `2.` might be sufficient.

In addition, when `runtime_cxxmodules` is ON, all tests supports *also* (a) **\*after\*** a second build directory configuration as their library is added indirectly to the `all` target by being added to the `module_idx` target (This is likely unintentional and explain why the problem solved by this PR was not seen in local development; it is not clear why it was not seen in the jenkings PR testing that introduced the test).

If we resolved to only use/support the implementation (a) we would gain by the removal of a set of `spurrious` tests (eg `roottest-root-io-stdpair-collection-build`) and thus likely reducing `ctest` run-time) but would lose support for pattern `2.`

### DIFF
--- a/root/io/stdpair/CMakeLists.txt
+++ b/root/io/stdpair/CMakeLists.txt
@@ -4,10 +4,10 @@ set(CMAKE_ROOTTEST_NOROOTMAP ON)
 set(RootExternalIncludes -e "gInterpreter->AddIncludePath(\"-I${CMAKE_CURRENT_SOURCE_DIR}\");")
 
 ROOT_GENERATE_DICTIONARY(G__CmsPairCollection ${CMAKE_CURRENT_SOURCE_DIR}/cmspair.h LINKDEF FullLinkDef.h)
-ROOTTEST_LINKER_LIBRARY(CmsPairCollection G__CmsPairCollection.cxx LIBRARIES ${ROOT_LIBRARIES})
+ROOTTEST_LINKER_LIBRARY(CmsPairCollection TEST G__CmsPairCollection.cxx LIBRARIES ${ROOT_LIBRARIES})
 
 ROOT_GENERATE_DICTIONARY(G__CmsPair ${CMAKE_CURRENT_SOURCE_DIR}/cmspair.h LINKDEF PairLinkDef.h)
-ROOTTEST_LINKER_LIBRARY(CmsPair G__CmsPair.cxx LIBRARIES ${ROOT_LIBRARIES})
+ROOTTEST_LINKER_LIBRARY(CmsPair TEST G__CmsPair.cxx LIBRARIES ${ROOT_LIBRARIES})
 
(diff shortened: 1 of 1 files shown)


hahnjo:
I think we should only do one of the two things here; as the test is currently using b, I'd favor this approach. The current setup leads to the library being built twice, once via `make` and then another time via `ctest` (because of `${always-make}`).

> it is not clear why it was not seen in the jenkings PR testing that introduced the test

Because the PR testing uses Ninja, only the nightlies (on macOS) use `make`. Yes, that's bad, I've raised this when investigating the original problem fixed in https://github.com/root-project/roottest/pull/858

pcanal:
On the other hand, @bellenot wouldn't moving the linking from the `ctest` step to the `ROOT build` step alleviate/solve problems like the one addressed in https://github.com/root-project/roottest/pull/876 ?

pcanal:
@hahnjo In __this__ test, the `--always-make` seems unnecessary.  Namely if I remove it and then touch a file in the roottest source directory, the library get rebuild.   Can you remind me the case where it was necessary?

pcanal:
Answering my own question, without--always-make` the local (direct?) dependencies (from roottest) are properly followed but changes in ROOT proper are ignored and there is no way to 'clean' (besides `rm the_path_to_files`). :(

root/io/stdpair/CMakeLists.txt
```diff
@@ -10,11 +10,11 @@ ROOT_GENERATE_DICTIONARY(G__CmsPair ${CMAKE_CURRENT_SOURCE_DIR}/cmspair.h LINKDE
 ROOTTEST_LINKER_LIBRARY(CmsPair TEST G__CmsPair.cxx LIBRARIES ${ROOT_LIBRARIES})
 
 ROOT_ADD_TEST(roottest-root-io-stdpair-collection-build 
-              COMMAND ${CMAKE_COMMAND} --build ${CMAKE_BINARY_DIR} --config $<CONFIG> --target CmsPairCollection${fast}
+              COMMAND ${CMAKE_COMMAND} --build ${CMAKE_BINARY_DIR} --config $<CONFIG> --target G__CmsPairCollection${fast}  CmsPairCollection${fast} -- ${always-make}
```
Axel-Naumann, line 13: Where is `${always-make}` being set? Is `-` an allowed character for a CMake variable name?

Axel-Naumann (approved):

PR #880, opened 2022-06-24
Disable runDeleteWarning test to avoid false positives

eguiraud:
This test relies on a use-after-delete (on purpose) to check that the
use-after-delete-detection logic in TList works properly.
Since the kNotDeleted mechanism relies on UB, sometimes it does not
work and this test segfaults instead.

### DIFF
--- a/root/collection/CMakeLists.txt
+++ b/root/collection/CMakeLists.txt
@@ -24,7 +24,11 @@ ROOTTEST_ADD_TEST(runTExMap
                   ERRREF runTExMap.eref )
 endif()
 
-ROOTTEST_ADD_TEST(runDeleteWarning
-                  MACRO runDeleteWarning.C
-                  OUTCNVCMD sed -e s:0x[0-9a-fA-F]*:0xRemoved:g
-                  OUTREF DeleteWarning.ref)
+# Disabled to avoid false positives.
+# This test relies on a use-after-delete (on purpose) to check that the use-after-delete-detection
+# logic in TList works properly. Since the kNotDeleted mechanism relies on UB, sometimes it does not work and this
+# test segfaults instead.
+#ROOTTEST_ADD_TEST(runDeleteWarning
+#                  MACRO runDeleteWarning.C
+#                  OUTCNVCMD sed -e s:0x[0-9a-fA-F]*:0xRemoved:g
+#                  OUTREF DeleteWarning.ref)


Axel-Naumann (approved):
But as discussed, @pcanal might want to discuss and revert that after the merge.

Axel-Naumann:
Random failure log: https://lcgapp-services.cern.ch/root-jenkins/view/ROOT%20Nightly/job/root-nightly-master/LABEL=ROOT-ubuntu2204,SPEC=soversion,V=master/3212/testReport/projectroot.roottest.root/collection/roottest_root_collection_runDeleteWarning/

eguiraud:
Can't the compiler elide the code that unsets the bit even if you invoke the destructor directly? It doesn't _have_ to execute it if it sees that there is no non-UB access possible to that bit afterwards

pcanal:
>  It doesn't have to execute it if it sees that there is no non-UB access possible to that bit afterwards

It might if it sees enough of the code (in which we need to add layers of indirection) but it should not because the affected address is part of a collection.

eguiraud:
I don't understand this last comment, the compiler definitely sees enough in the test in question:

```cpp
auto o = new TObject();
auto l = new TList();
l->SetName("my own list");
l->Add(o);
delete o; // compiler can elide the bit unsetting, because...
l->Clear(); // ...accessing `o` here is clearly UB
```

pcanal:
> delete o; // compiler can elide the bit unsetting, because...

But how?  If I am not mistaken the bit is set in '~TObject` at the end:

```
   fBits &= ~kNotDeleted;

   if (fgObjectStat && gObjectTable) gObjectTable->RemoveQuietly(this);
}
```
And how would the compiler 'know' that 
(a) this is called from operator delete
(b) the bit is not used in `RemoveQuietly`
?

And I was exploring
```
auto o = new TObject();
auto l = new TList();
l->SetName("my own list");
l->Add(o);
o::~TObject();
l->Clear();
```
but obviously did not close the loop (Possibly because I was still pondering why it would crash anyway .. maybe)

eguiraud:
Ah, you are right that the TU boundary might prevent the compiler to do the kind of optimization I mentioned.

Taking a look with `gdb` it looks like `kNotDeleted` is reset, but then those bits are immediately modified again by `_int_free`:

```cpp
#0  tcache_put (tc_idx=0, chunk=0x55555562e400) at malloc.c:3183
#1  _int_free (av=0x7ffff4dfdbc0 <main_arena>, p=0x55555562e400, have_lock=0) at malloc.c:4481
#2  0x00007ffff4c9c8f3 in __GI___libc_free (mem=<optimized out>) at malloc.c:3391
#3  0x00007ffff7b30e64 in TStorage::ObjectDealloc (vp=0x55555562e410) at ../core/base/src/TStorage.cxx:362
#4  0x00007ffff7b14518 in TObject::operator delete (ptr=0x55555562e410) at ../core/base/src/TObject.cxx:1001
#5  0x00007ffff7b11cba in TObject::~TObject (this=0x55555562e410, __in_chrg=<optimized out>) at ../core/base/src/TObject.cxx:91
#6  0x0000555555556242 in main () at foo.cpp:9
```

where the line that accidentally sets the bit again is 3181 here:

```cpp
 3172  /* Caller must ensure that we know tc_idx is valid and there's room
 3173     for more chunks.  */
 3174  static __always_inline void
 3175  tcache_put (mchunkptr chunk, size_t tc_idx)
 3176  {
 3177    tcache_entry *e = (tcache_entry *) chunk2mem (chunk);
 3178
 3179    /* Mark this chunk as "in the tcache" so the test in _int_free will
 3180       detect a double free.  */
 3181    e->key = tcache_key;
 3182
 3183    e->next = PROTECT_PTR (&e->next, tcache->entries[tc_idx]);
```

and this version of the test circumvents that behavior of `_int_free` so it does not crash (at least on my laptop):

```cpp
  void *mem = malloc(sizeof(TObject));
  auto o = new (mem) TObject();
  auto l = new TList();
  l->SetName("my own list");
  l->Add(o);
  o->~TObject();
  l->Clear();
  free(mem);
```

If my understanding is correct, this also means that the test failure is real in the sense that the `kNotDeleted` mechanism does not correctly work on platforms where `free` has that behavior.

pcanal:
> If my understanding is correct, this also means that the test failure is real in the sense that the kNotDeleted mechanism does not correctly work on platforms where free has that behavior.

And for better or worse, it also means that the test in its current form is "as intended" and detected what we hoped it detected ...

eguiraud:
At least Arch Linux and Ubuntu 22.04 are broken, see also https://lcgapp-services.cern.ch/root-jenkins/view/ROOT%20Nightly/job/root-nightly-master/LABEL=ROOT-ubuntu2204,SPEC=soversion,V=master/3214/testReport/junit/projectroot.roottest.root/collection/roottest_root_collection_runDeleteWarning

Axel-Naumann:
@pcanal can you remind us whether we need this for more than checking whether something is accessing deleted memory, i.e. "better diagnostics" than a crash?

pcanal:
> i.e. "better diagnostics" than a crash?

I think you meant "better ... than random behavior" :) 

It is used in the collection to avoid inadvertent delete memory use or double delete.  It has been helpful in particular because of the indirect ownerships.   On architecture that salt the memory so that they can (already) have (better/some) diagnostics, we can probably skip this test (i.e. we probably have no alternative).

Axel-Naumann:
> It is used in the collection to avoid inadvertent delete memory use or double delete.

This suggests that it's a functionality that the collections depends on, rather than a quality of life improvement (diagnostic instead of crash). So - does ROOT rely on this or can we do without (if need be)?

eguiraud:
Ping @pcanal

pcanal:
We (I) need to extend the core library to detect when there is a memory checker (or similar) that leads to the memory being salted after a delete and in that case disable our own already-deleted test (since it is now ineffective anyway and assumingly it is already doing the job of warning the user about use-after-delete).  And subsequently the test needs to also be disabled in those cases.

eguiraud:
Cool, then this PR can be closed. I opened an issue instead: https://github.com/root-project/root/issues/11330

PR #881, opened 2022-06-24
Disable runDeleteWarning test to avoid false positives (v6.26 backport)

eguiraud:
This test relies on a use-after-delete (on purpose) to check that the
use-after-delete-detection logic in TList works properly.
Since the kNotDeleted mechanism relies on UB, sometimes it does not
work and this test segfaults instead.


pcanal:
How does it crash?

PR #884, opened 2022-07-07
Fix MemberComments test after modify TH1.h header

linev:
Should fix changes after https://github.com/root-project/root/pull/10915


bellenot (approved):

linev:
All runMemberComments failures should be fixed after  https://github.com/root-project/roottest/pull/884

PR #885, opened 2022-07-10
[DF] Add test for friend alignment

vepadulano:
Sibling PR of https://github.com/root-project/root/pull/10939

### DIFF
--- a/python/distrdf/dask/check_friend_trees_alignment.py
+++ b/python/distrdf/dask/check_friend_trees_alignment.py
@@ -1,0 +1,79 @@
+import os
+
+import pytest
+
+import ROOT
+from DistRDF.Backends import Dask
+
+
+TREENAMES = [f"distrdf_check_friend_trees_alignment_dask_tree_{i}.root" for i in range(1, 7)]
+FILENAMES = [f"distrdf_check_friend_trees_alignment_dask_file_{i}.root" for i in range(1, 7)]
+
+
+def create_dataset():
+    df = ROOT.RDataFrame(60).Define("x", "rdfentry_")
+
+    range_limits = list(range(0, 61, 10))
+    opts = ROOT.RDF.RSnapshotOptions()
+    opts.fAutoFlush = 1
+    for idx, (begin, end) in enumerate(zip(range_limits, range_limits[1:])):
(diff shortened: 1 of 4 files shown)


python/distrdf/dask/check_friend_trees_alignment.py
```diff
@@ -0,0 +1,77 @@
...
+
+TREENAMES = [f"distrdf_check_friend_trees_alignment_dask_tree_{i}.root" for i in range(1, 7)]
+FILENAMES = [f"distrdf_check_friend_trees_alignment_dask_file_{i}.root" for i in range(1, 7)]
+
+
+def create_dataset():
+    df = ROOT.RDataFrame(60).Define("x", "rdfentry_")
+
+    ranges = list(range(0, 61, 10))
+    for i in range(len(ranges[:-1])):
```
eguiraud, line 17: I think the python idiom for something like this is:

```python
for begin, end in zip(ranges, ranges[1:]):
```

python/distrdf/dask/check_friend_trees_alignment.py
```diff
@@ -0,0 +1,77 @@
...
+    yield
+    for name in FILENAMES:
+        os.remove(name)
+
+
+@pytest.mark.usefixtures("setup_testfriendtreesalignment")
+class TestDaskFriendTreesAlignment:
+    """Check that distrdf processes the correct entries from friend trees"""
+
+    def test_sum_friend_column(self, connection):
```
eguiraud, line 55: for my education, where does `connection` come from?

eguiraud (approved):
Thank you Vincenzo!

It's sad that we have to duplicate the whole test for spark and dask :/

python/distrdf/dask/check_friend_trees_alignment.py
vepadulano: `connection` is a pytest fixture, creates a single Dask/Spark connection object for all the tests in the Dask/Spark suite

vepadulano: They are defined in the respective folders, for example the Dask one is at https://github.com/root-project/roottest/blob/master/python/distrdf/dask/conftest.py

PR #886, opened 2022-07-23
[tree] Test TEntryList::GetEntryAndTree with index larger than max int

vepadulano:
Regression test for https://github.com/root-project/root/issues/11026

Sibling PR of https://github.com/root-project/root/pull/11028


root/tree/entrylist/regression_intoverflow.cxx
```diff
@@ -0,0 +1,80 @@
...
+higher than the maximum value of int. The chain is connected in lockstep with a
+TEntryList which has one sub TEntryList per tree in the chain.
+
+The test finally calls `TEntryList::GetEntryAndTree` for the entry index right
+after the maximum int value. This triggered the int overflow bug described in
+the linked issue.
+*/
+TEST(TEntryList, EntryGreaterThanMaxInt)
+{
+    constexpr int limit = std::numeric_limits<int>::max();
```
jalopezg-r00t, line 42: Should not rely on transitive inclusion and we should perhaps add `#include <limits>`.

jalopezg-r00t (approved):
LGTM (but see comment)!

root/tree/entrylist/regression_intoverflow.cxx
vepadulano: Good catch, I had it in the original reproducer and somehow missed to copy it :(

PR #891, opened 2022-07-27
[DF] Check C++ exception is correctly propagated in Dask 

vepadulano:
Test that C++ runtime_error is propagated correctly as a Python
RuntimeError with the original message from RDF. This is only needed in
Dask, Spark does not try to serialize the error instance.

More details at https://github.com/root-project/root/issues/11050


vepadulano:
Sibling PR of https://github.com/root-project/root/pull/11047

etejedor (approved):

PR #892, opened 2022-08-04
[DF] Revert "[DF] Update test reference"

eguiraud:
This reverts commit e099f75e58dd039b8e3d35ce22ee5f389dd2900c.
We now keep the original ordering of the input branches in
Snapshot.


pcanal (approved):

eguiraud:
the `datasource_root` failure is unrelated although slightly worrying (although the ROOT datasource is not actually used anywhere and it's internal -- it's just there as a sanity check for the RDataSource API).

PR #894, opened 2022-08-09
[RF] Update `testNestedPDFs` to avoid integrating a RooRealVar

guitargeek:
Related to https://github.com/root-project/root/pull/11162.

### DIFF
--- a/root/roofitstats/vectorisedPDFs/testNestedPDFs.cxx
+++ b/root/roofitstats/vectorisedPDFs/testNestedPDFs.cxx
@@ -40,9 +40,10 @@ class TestNestedPDFs : public PDFTest
       auto a2 = new RooRealVar("a2", "a2", 1., 1., 5.);
       a0->setConstant(true);
       a1->setConstant(true);
+      auto xId = new RooProduct("xId", "x", RooArgList(*x));
       auto xSq = new RooProduct("xSq", "x^2", RooArgList(*x, *x));
       auto one = new RooConstVar("one", "one", 1.);
-      auto pol = new RooRealSumPdf("pol", "pol", RooArgList(*one, *x, *xSq), RooArgList(*a0, *a1, *a2));
+      auto pol = new RooRealSumPdf("pol", "pol", RooArgList(*one, *xId, *xSq), RooArgList(*a0, *a1, *a2));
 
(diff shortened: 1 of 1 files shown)


root/roofitstats/vectorisedPDFs/testNestedPDFs.cxx
```diff
@@ -40,9 +40,10 @@ class TestNestedPDFs : public PDFTest
       auto a2 = new RooRealVar("a2", "a2", 1., 1., 5.);
       a0->setConstant(true);
       a1->setConstant(true);
+      auto xId = new RooProduct("xId", "x", RooArgList(*x));
       auto xSq = new RooProduct("xSq", "x^2", RooArgList(*x, *x));
       auto one = new RooConstVar("one", "one", 1.);
-      auto pol = new RooRealSumPdf("pol", "pol", RooArgList(*one, *x, *xSq), RooArgList(*a0, *a1, *a2));
+      auto pol = new RooRealSumPdf("pol", "pol", RooArgList(*one, *xId, *xSq), RooArgList(*a0, *a1, *a2));
```
lmoneta, line 46: It is not clear to me why it is needed to create a dummy RooProduct. 
Can not the RooRealVar used as a RooAbsReal as a simple function: `f(x) = x` and what the advantages of using RooProduct

lmoneta (approved):

root/roofitstats/vectorisedPDFs/testNestedPDFs.cxx
guitargeek: You can't use a RooRealVar as the identity function directly, because the integration is implemented wrong.

So I had two options:
* fix the integration of RooRealVars
* disallow the integration of RooRealVars

As RooRealVars are apparently not used as identity functions anywhere in RooFit outside of artificial unit tests, I concluded there is not any demand for integrating RooRealVars, also because someone would have seen the bug before. Therefore, the quickest solution is to disallow the integration.

As for the RooProduct, one could have used also used anything else that can represent an identity, like RooAddition or RooFormulaVar. But I went for RooProduct in this unit test, because it is already used for x squared.

guitargeek: More details in the other PR: https://github.com/root-project/root/pull/11162

PR #897, opened 2022-09-09
Update runDeleteWarning (and iobug.C)

pcanal:


root/meta/MemberComments.ref
```diff
@@ -1,5 +1,5 @@
 
-Processing runMemberComments.C...
+Processing /home/pcanal/root_working/code/root/roottest/root/meta/runMemberComments.C...
```
Axel-Naumann, line 2: That should not be necessary

Axel-Naumann (approved):
But see comment

PR #902, opened 2022-09-22
macOS: remove "-undefined dynamic_lookup" from linker:

Axel-Naumann:
Seems not needed, and it causes the warning
ld: warning: -undefined dynamic_lookup may not work with chained fixups


Axel-Naumann:
Windows failure is unrelated. @pcanal this is ready for review

pcanal (approved):
Thanks

PR #909, opened 2022-10-19
[RF] Disable Python tests using RooFit if no RooFit is available

guitargeek:
The tests where RooFit objects are used to test something general like
pickling are skipped if RooFit was not built. This is checked via
CMake and passed to the test as an environment variable, in the same way
as CMake checks for PyROOT legacy.

The tests that are testing actual RooFit pythonizations are moved to the
main ROOT repository because that's where the other RooFit pythonization
tests are.

Closes https://github.com/root-project/root/issues/11605.


guitargeek:
@phsft-bot build just on ROOT-performance-centos8-multicore/default with flags -DCTEST_TEST_EXCLUDE_NONE=On -Droofit=OFF

python/pickle/PyROOT_readtests.py
```diff
@@ -134,7 +134,10 @@ def test5ReadCustomTypes( self ):
 
    def test6ReadCustomTypes( self ):
       """[ROOT-10810] Test reading a RooDataSet with weights"""
-      ROOT.gEnv.SetValue("RooFit.Banner", 0)
```
etejedor, line 137: Why is this removed?

python/pickle/PyROOT_readtests.py
```diff
@@ -134,7 +134,10 @@ def test5ReadCustomTypes( self ):
...
+
+      if os.environ.get('ROOFIT') == 'False':
```
etejedor, line 138: What is being unpickled in this test are Roofit objects?

python/pickle/PyROOT_readtests.py
guitargeek: The banner got completely removed: https://github.com/root-project/root/pull/9965

python/pickle/PyROOT_readtests.py
guitargeek: Taking a look at the `writetests` file, it is because there was a Jira issue on pickling RooDataSets, which is why it's tested here to read them back in

python/pickle/PyROOT_readtests.py
etejedor: Deserves a separate commit, then? :)

guitargeek: Alright, I'll split them up :+1:

guitargeek:
@phsft-bot build just on ROOT-performance-centos8-multicore/default with flags -DCTEST_TEST_EXCLUDE_NONE=On -Droofit=OFF

PR #912, opened 2022-11-12
macOS: remove "-undefined dynamic_lookup" from linker:

Axel-Naumann:
Seems not needed, and it causes the warning
ld: warning: -undefined dynamic_lookup may not work with chained fixups

Backport of #902


Axel-Naumann:
This actually fixes the mac11 error - Windows is spurious and the warnings are pre-existing.

PR #922, opened 2022-12-01
Add tests for Stats and StdDev in distributed RDataFrame

vepadulano:
Sibling PR of https://github.com/root-project/root/pull/11815

### DIFF
--- a/python/distrdf/dask/check_reducer_merge.py
+++ b/python/distrdf/dask/check_reducer_merge.py
@@ -352,6 +352,46 @@ class TestReducerMerge:
 
         assert std.GetValue() == pytest.approx(1, 0.01), f"{std.GetValue()}"
 
+    def test_distributed_stats(self, connection):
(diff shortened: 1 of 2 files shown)


python/distrdf/dask/check_reducer_merge.py
```diff
@@ -341,6 +341,57 @@ def test_redefine_one_column(self, connection):
         assert sum_before.GetValue() == 10.0
         assert sum_after.GetValue() == 20.0
 
+    def test_distributed_stddev(self, connection):
+        """Test support for the StdDev action."""
+        df = (
+            Dask.RDataFrame(10_000, daskclient=connection)
+                .Define("v", "gRandom->Gaus()")
```
eguiraud, line 348: here and below it would be more robust to use a fixed series of numbers rather than sampling them

eguiraud (approved):
LGTM

PR #923, opened 2022-12-06
[DF][NFC] Remove old tdfentry_ in tests

vepadulano:


vepadulano:
Following the discussion at https://github.com/root-project/root/pull/11843 , there are no changes to be made in roottest, we want to keep the `rdfentry_` special column usable in the empty source case. This PR now simply changes the old spelling `tdfentry_` still present in some tests to the new one.

eguiraud (approved):

PR #927, opened 2022-12-15
Test gDirectory clean and TContext race condition

pcanal:
Add test that the thread local gDirectory and TContext are properly updated if another thread deletes a TFile there are pointing to.

Also attempt to test the race condition between the creation of the TContext and the gDirectory update (to reflect a deleted TFile). However it is quite rare (the window of miss behavior is small, i.e. within the TContext constructor).

See https://github.com/root-project/root/pull/11908


pcanal:
@phsft-bot build

Axel-Naumann (approved):

PR #928, opened 2022-12-21
remove KahanSum implicit conversion

egpbos:
`KahanSum` objects will no longer be implicitly converted to their underlying type after https://github.com/root-project/root/pull/11941. This commit explicitly converts them.


guitargeek (approved):
Thanks!

PR #932, opened 2023-01-18
Replace sprintf in macros

hahnjo:
There are still many other occurrences, but these changes should be sufficient to make the tests pass on macOS 13.


hahnjo:
@phsft-bot build just on mac13/default, windows10/cxx14

hahnjo:
@phsft-bot build just on mac13/default

PR #936, opened 2023-02-20
Revert "[DF] Disable distributed tests on M1 ARM"

hahnjo:
This reverts commit 3971a5488667a8d1a5d7a52228bc56cfe735ef2f.

Closes https://github.com/root-project/root/issues/10061

### DIFF
--- a/python/distrdf/CMakeLists.txt
+++ b/python/distrdf/CMakeLists.txt
@@ -4,8 +4,7 @@
 # * The Python version is greater than 3.7
 if (ROOT_dataframe_FOUND AND
     ROOT_pyroot_FOUND AND
-    PYTHON_VERSION_STRING_Development_Main VERSION_GREATER_EQUAL 3.7
-    AND NOT (APPLE AND CMAKE_SYSTEM_PROCESSOR MATCHES arm64))
+    PYTHON_VERSION_STRING_Development_Main VERSION_GREATER_EQUAL 3.7)
 
     file(COPY test_headers DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
     file(COPY test_shared_libraries DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
--- a/python/distrdf/dask/check_backend.py
+++ b/python/distrdf/dask/check_backend.py
@@ -1,4 +1,5 @@
 import os
+import platform
 import pytest
 
 import DistRDF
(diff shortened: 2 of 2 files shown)


hahnjo:
@phsft-bot build just on mac11arm/default, mac12arm/default, mac13arm/default

hahnjo:
Ok, mac11arm and mac12arm didn't run the tests, but mac13arm did; I'm a bit confused by the failure in https://lcgapp-services.cern.ch/root-jenkins/job/roottest-pullrequests-build/12233/testReport/projectroot.roottest.python.distrdf/dask/roottest_python_distrdf_dask_test_all/ - that looks like the test is throwing an exception somewhere which we know doesn't work in M1. @vepadulano would it make sense to only disable that one (part of the) test?

vepadulano:
Yes that test precisely throws an exception on bad usage of `Graph` so if that doesn't work we should disable it.

hahnjo:
> Yes that test precisely throws an exception on bad usage of `Graph` so if that doesn't work we should disable it.

Maybe stupid question: How do we do that? Can we check for Apple & AArch64 and then disable the tests involving exceptions? Do we already have an existing test that does this, from which I could copy?

hahnjo:
@phsft-bot build just on mac11arm/default, mac12arm/default, mac13arm/default with flags -DCTEST_TEST_EXCLUDE_NONE=ON

hahnjo:
@phsft-bot build just on mac12arm/default, mac13arm/default with flags -DCTEST_TEST_EXCLUDE_NONE=ON

vepadulano:
> Build failed on mac13arm/default. Running on 194.12.149.74:/Users/sftnight/build/workspace/roottest-pullrequests-build [See console output](https://lcgapp-services.cern.ch/root-jenkins/job/roottest-pullrequests-build/13251/console).
> ### Failing tests:
> 
>     * [projectroot.roottest.python.distrdf.dask.roottest_python_distrdf_dask_test_all](https://lcgapp-services.cern.ch/root-jenkins/job/roottest-pullrequests-build/13251/testReport/projectroot.roottest.python.distrdf/dask/roottest_python_distrdf_dask_test_all/)

I have started to see this failure on DistRDF + Dask tests today for the first time. But it seems it's not strictly related to this PR (or MacOS in general), e.g. https://github.com/root-project/root/pull/13912#issuecomment-1773534878 or https://github.com/root-project/root/runs/17955813786 .

Investigating :+1:

hahnjo:
I'm not sure it's the same:
```
libc++abi: terminating due to uncaught exception of type std::runtime_error: Graph was applied to a mix of scalar values and collections. This is not supported.
```
is a C++ exception which we cannot catch on Apple Silicon. I already described this in https://github.com/root-project/roottest/pull/936#issuecomment-1436999157 and I continue to think we need to disable this test.

The other two are `ZeroDivisionError: division by zero`, which I also don't see in the link above...

hahnjo:
@phsft-bot build just on mac13arm/default with flags -DCTEST_TEST_EXCLUDE_NONE=ON

vepadulano (approved):
Indeed this needs to be disabled anyway as you say, I believe using the pytest interface is more appropriate. Another issue appeared this week with the same test, but it's not due to the same problem (see [this PR](https://github.com/root-project/root/pull/13926)) and is already fixed.

PR #939, opened 2023-03-16
 Add test for #12378 (typedef not found first try)

pcanal:


root/meta/autoloading/ROOT-12378/classes00.hh
jalopezg-git, line 1: Why do we need these `classesXX.hh` files (instead of just directly referring to the `#include`d file)?

root/meta/autoloading/ROOT-12378/Element.hh
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+template <typename T> struct TestElementLink {};
+
+
```
jalopezg-git, line 5: Remove trailing empty lines.
```suggestion
```

root/meta/autoloading/ROOT-12378/exec_testtypedef.C
```diff
@@ -0,0 +1,23 @@
...
+     {
+            TInterpreter::SuspendAutoLoadingRAII autoloadOff(gInterpreter);
+            TClassEdit::GetNormalizedName(normalizedName, name);
+     }
+     cerr << "Normalized named is  : " << normalizedName << '\n';
+     cerr << "Second attempt: " << TClass::GetClass(name.c_str())
+          << '\n';
+     return 1;
+  } else {
+     return 0;
```
jalopezg-git, line 21: Maybe reverse the `if` block to make the test more readable, as in:
```c++
if (cl)
   return 0;
// [rest of the code here]
```

jalopezg-git (commented):
I have added a few comments, but I'm missing information.  In particular, I assume that the issue is still present in the `master` branch; thus, merging this will make this test fail in all CI configurations (?)

root/meta/autoloading/ROOT-12378/classes00.hh
pcanal: Actually I confused myself.  This version uses `genreflex` so those headers are necessary.

root/meta/autoloading/ROOT-12378/exec_testtypedef.C
pcanal: Note early return is 'broken' for unnamed script so apply 'strictly' the proposed change fails: it still executes the lines following the if+return statement (and sometimes skip the next statement).

jalopezg-git (approved):
LGTM, but I would also like to take a look at https://github.com/root-project/root/pull/12488...

PR #940, opened 2023-03-20
Replace sprintf in macros

hahnjo:
There are still many other occurrences, but these changes should be sufficient to make the tests pass on macOS 13.

(cherry picked from commit 791ddcf530ce7e76b2411328ac25ad45173dfb71)


hahnjo:
@phsft-bot build only on mac11arm/default, mac12arm/default, mac13arm/default

hahnjo:
@phsft-bot build just on mac11arm/default, mac12arm/default, mac13arm/default

PR #943, opened 2023-03-22
Add test for #12358 (tuple and unloading)

pcanal:


root/meta/tuple/selection.xml
Axel-Naumann, line 1: This file seems to not be needed?

Axel-Naumann (approved):

pcanal:
The `mac11/cxx14` failure seems unrelated (timeout in tests not involving `TSelectorDraw`).

PR #944, opened 2023-03-23
Fix roottest being configured without pyroot

hageboeck:
Accidentally, I configured root without pyroot. This leads to a few test failures because of missing `if`s in roottest's cmake config.


hageboeck:
@phsft-bot build

bellenot (approved):
LGTM

PR #946, opened 2023-03-27
Test for race condition between dlopen and TClassTable ops.

pcanal:
This is testing the ROOT issue https://github.com/root-project/root/issues/12552


bellenot (commented):
`exec_dlopenrace.C` uses posix functions, so it's not going to work on Windows, unless you (or someone else) make it cross-platform...

pcanal:
>  uses posix functions, so it's not going to work on Windows

Yes, the test was meant to be disable on Windows :(

root/meta/race-12552/exec_dlopenrace.C
```diff
@@ -0,0 +1,47 @@
...
+void repeatlookup(int n)
+{
+   static DictFuncPtr_t value = nullptr;
+
+   for(int i = 0; i < n; ++i) {
+     auto r = TClassTable::GetDictNorm("UserClass");
+     if (!value && r)
+       value = r;
+     // std::cerr << "Value of dictionary pointer has from " << (void*)value << " to " << (void*)r << '\n';
+     if (r != nullptr && r != value)
```
jalopezg-git, line 32: Why not simply `if (r != value)`?  Is it allowed to change back to `nullptr` even if we had a value before?

root/meta/race-12552/exec_dlopenrace.C
```diff
@@ -0,0 +1,47 @@
...
+     auto r = dlopen("./libUser.so", RTLD_LAZY);
+     if (!r) {
+        std::cerr << "Failed to open libUser.so at iteration " << i << "\n";
+        return;
+     }
+     // Waste some time
+     for(int j = 0; j < 10; ++j)
+       auto d = TClassTable::GetDictNorm("UserClass");
+     // std::cerr << "Seeing dict pointer: " << (void*)d << '\n';
+     dlclose(r);
```
jalopezg-git, line 19: This uses the platform-specific `dlopen()` and `dlclose()`.  Can't we make it platform-independent?  As @bellenot mentioned, in its current state, the test will not build on Windows.

Alternatively, the test should be disabled on non-POSIX platforms.

pcanal: It is now disable on WIndows.  I elect not to run on Windows as I can use the ROOT platform independent interface (they do too much) and running/testing on one platform is sufficient to test this race condition.

root/meta/race-12552/exec_dlopenrace.C
pcanal: `r == nullptr && r != value` happens and is 'expected' in this test in the case where `value` has already been assigned and `TClassTable::GetDictNorm` is called between the invocation of `dclose` and the invocation of `dlopen` (i.e. at a point in time where the library is unloaded).

root/meta/race-12552/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+ROOTTEST_GENERATE_REFLEX_DICTIONARY(libUser User.hpp SELECTION User.xml LIBNAME libUser NO_ROOTMAP)
+set(depends ${GENERATE_REFLEX_TEST})
+
+if(NOT MSVC)
+   ROOTTEST_ADD_AUTOMACROS(DEPENDS ${depends})
```
jalopezg-git, line 5: If the test is to be disabled on Windows, I think `ROOTTEST_GENERATE_REFLEX_DICTIONARY()` in line 1 can also be skipped, right?  That should save some cycles.

pcanal:
The problem is a (real) quirk of the test.  It linked to the repeated unload and loading of the same library in parallel.  During the unloading and loading cycle the TClass object representing the user classes (from that library) are created and deleted and it seems that the pointer value is captured and used outside the main lock (and this is done concurrently we the deletion).  This needs to be fixed but is unlikely to happen in the wild.

pcanal:
Tracking of the issue at https://github.com/root-project/root/issues/12715

PR #947, opened 2023-03-29
Remove test for std::span:

Axel-Naumann:
Fixes C++20 builds.

We can rely on the stdlibs to provide a working version. Our backport is frozen; we support what we support, which is less than what is tested here, as can be seen by the failures appearing when running this test with std::span instead of our backport.

I could maybe fix our backport but that seems not a good investment.


pcanal (approved):

pcanal:
How does the test fail with C++20?  (I.e. any chances that the current code (outside this test) that uses std::span is using features that are "similar" to this test)

Axel-Naumann:
Diagnostics (now also part of the commit log):
```
In file included from input_line_24:1:
/private/tmp/workspace/src/roottest/root/collection/arrayViewBasic.C:77:5: error: no matching function for call to 'show_int_array'
    show_int_array<std::span<const int>>({1, 2, 3, 4}); //Note initializer_list cannot contain non-const types, but span can.
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/private/tmp/workspace/src/roottest/root/collection/arrayViewBasic.C:59:6: note: candidate function template not viable: cannot convert initializer list argument to 'std::span<const int, 18446744073709551615ull>'
void show_int_array(T view)
     ^
/private/tmp/workspace/src/roottest/root/collection/arrayViewBasic.C:29:31: error: no member named 'at' in 'std::span<int, 18446744073709551615ull>'
            std::cout << view.at(i);
                         ~~~~ ^
/private/tmp/workspace/src/roottest/root/collection/arrayViewBasic.C:61:4: note: in instantiation of function template specialization 'show_int_array_at<std::span<int, 18446744073709551615ull> >' requested here
   show_int_array_at(view);
   ^
/private/tmp/workspace/src/roottest/root/collection/arrayViewBasic.C:74:5: note: in instantiation of function template specialization 'show_int_array<std::span<int, 18446744073709551615ull> >' requested here
    show_int_array<std::span<int>>(good_old_c_array);
    ^
/private/tmp/workspace/src/roottest/root/collection/arrayViewBasic.C:29:31: error: no member named 'at' in 'std::span<const int, 18446744073709551615ull>'
            std::cout << view.at(i);
                         ~~~~ ^
/private/tmp/workspace/src/roottest/root/collection/arrayViewBasic.C:61:4: note: in instantiation of function template specialization 'show_int_array_at<std::span<const int, 18446744073709551615ull> >' requested here
   show_int_array_at(view);
   ^
/private/tmp/workspace/src/roottest/root/collection/arrayViewBasic.C:76:5: note: in instantiation of function template specialization 'show_int_array<std::span<const int, 18446744073709551615ull> >' requested here
    show_int_array<std::span<const int>>(vector);
    ^
```

pcanal:
I think that is fine to ignore those.  `at` can be replace with `operator[]` and still work with both our `span` and `C++20`'s `span`.  The initializer list is more annoying (probably hard to write code compatible with both) but should be rare (both in actual use and in the type that might cause problem (i.e. maybe it is just `const` something ).

Axel-Naumann:
You cannot list-init a `std::span` (for now / C++20).

PR #951, opened 2023-04-19
[core][DF] Fix `-Wreturn-type` in tests

jalopezg-git:
As of PR https://github.com/root-project/root/pull/12654, returning a non-`void` expression from `void`-returning function is only valid in an unnamed macro (in which case the diagnostic is filtered).


pcanal (approved):
LGTM

eguiraud (approved):
Thanks!

PR #954, opened 2023-04-28
dictSelection: ignore data members inside STL.

pcanal:
Newer implementation have the members in the class itself rather than a base class

### DIFF
--- a/root/meta/dictSelection/execAtlasTest2.C
+++ b/root/meta/dictSelection/execAtlasTest2.C
@@ -23,6 +23,13 @@ void printMembers( TClass* cl ) {
 
    std::cout << "Members of class " << cl->GetName() << std::endl;
 
+   if (cl->GetCollectionProxy())
+      // Members does not matter (and some implementation have them in the base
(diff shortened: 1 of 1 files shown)


Axel-Naumann (approved):
The super annoying test failure is unrelated.

Axel-Naumann:
Merged manually into v6-28-00-patches as 4641ea1946ba349c237183f4034f5828e44a17b5

hahnjo:
After this PR it seems that `execAtlasTest2_WILL_FAIL` is red. @bellenot does this mean we need to remove the expectation that the test will fail on Windows?

bellenot:
> After this PR it seems that `execAtlasTest2_WILL_FAIL` is red. @bellenot does this mean we need to remove the expectation that the test will fail on Windows?

Most probably. Already working on it :wink:

Axel-Naumann:
> The super annoying test failure is unrelated.

And indeed I should have insisted with Jenkins to provide results for Windows. Sorry, @hahnjo - I am causing quite some problems lately 😞

PR #955, opened 2023-05-02
Custom allocator fixes

hahnjo:
Required to pass tests with newer versions of `libstdc++`.

### DIFF
--- a/root/collection/execMissing.C
+++ b/root/collection/execMissing.C
@@ -4,6 +4,11 @@
 template <typename T>
 class CustomAlloc : public std::allocator<T>
 {
+public:
+   template <typename U>
+   struct rebind {
+      using other = CustomAlloc<U>;
+   };
 };
 
 class Track {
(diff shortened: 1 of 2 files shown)


Axel-Naumann (approved):
LGTM thanks! (Could've been nice to mention the motivation for the change in the second commit's log, but we'll survive.)

hahnjo:
> Could've been nice to mention the motivation for the change in the second commit's log, but we'll survive.

I can add that, at least the part that I understand...

jalopezg-git (commented):
Thanks, @hahnjo :slightly_smiling_face:!  Don't forget to run `clang-format` on this patch -- there seems to be some code style inconsistency.

hahnjo:
> Don't forget to run `clang-format` on this patch -- there seems to be some code style inconsistency.

Done. At least partially, `clang-format` wanted to add another space before `public:` which is not our code style...

PR #957, opened 2023-05-03
Fix newClassDef tests with C++20

hahnjo:
C++20 made it invalid to use simple-template-ids for constructors and destructors: https://eel.is/c++draft/diff.cpp17.class#2
GCC 11 and later throw an error on this, with the unhelpful message `expected unqualified-id before '<some>' token`.


jalopezg-git (approved):
LGTM!  Many thanks!

PR #960, opened 2023-05-03
Automatically add missing `override` to root/tree and root/treeformula tests 

guitargeek:
This is done using `clang-tidy`.


root/tree/addresses/ConfigRecord.h
```diff
@@ -9,7 +9,7 @@ class ConfigRecord : public RecDataRecord<RecHeader>
 public:
   ConfigRecord() {}
   ConfigRecord(const RecHeader& header) : RecDataRecord<RecHeader>(header){} 
-  virtual ~ConfigRecord() {}
+  ~ConfigRecord() override {}
```
pcanal, line 12: I think you also need to change the `ClassDef` to `ClassDefOverride`

guitargeek: Thanks! I have updated the PR now. Took me some iterations because some classes needed to keep `ClassDef` because they didn't inherit from TObject.

guitargeek:
The final force push fixes `[projectroot.roottest.root.tree.evolution.roottest_root_tree_evolution_make](https://lcgapp-services.cern.ch/root-jenkins/job/roottest-pullrequests-build/12560/testReport/projectroot.roottest.root.tree/evolution/roottest_root_tree_evolution_make/)` that was failing before.

pcanal (approved):

PR #964, opened 2023-05-10
Revert "Test for race condition between dlopen and TClassTable ops."

hahnjo:
This test is quirky and known to fail in a couple of configurations, including the AddressSanitizer build, as reported in the original PR https://github.com/root-project/roottest/pull/946 and in the issue https://github.com/root-project/root/issues/12715.

This reverts commit 3186d71af69c67c51d9adec549edbce9fa42a6e3.


hahnjo:
ping @pcanal

pcanal:
The underlying problem is fixed by [#12863](https://github.com/root-project/root/pull/12863) and thus this reversion is no longer needed.

PR #968, opened 2023-05-25
dlopen race: speed up when dependent library are also unloaded.

pcanal:
To reduce the number of libraries unloaded on EL8 when clang++ (other platform don't seem to unload the dependent libraries) we preload some (but not all) of the libraries (keeping some to test the unloading and reloading of complex dictionary).

Also speed-up the test by stopping as soon as the lookups are done looping.

This is linked to https://github.com/root-project/root/pull/12863

### DIFF
--- a/root/meta/race-12552/User.hpp
+++ b/root/meta/race-12552/User.hpp
@@ -1,6 +1,13 @@
 #ifndef USER_CLASS_HPP
 #define USER_CLASS_HPP
 
-class UserClass {};
+#include <vector>
+#include <TString.h>
+
+struct UserClass {
+   // The dictionary for this class is generated using genreflex
+   // so the following is not enough to trigger a GenerateInitInstance
+   std::vector<TString> fValues;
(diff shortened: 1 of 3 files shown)


hahnjo (approved):
Note that this change on its own already "hides" the problem described in https://github.com/root-project/root/issues/12715

pcanal:
> Note that this change on its own already "hides" the problem described in https://github.com/root-project/root/issues/12715

Humm .. right .. and this is a problem per se ... The issue was being triggered by the unloading of `libHist` which is now no longer unloaded and thus we are no longer testing that code paths ... I need to update the code to add `libHist` back to the unload list .. or better yet (assuming it works out) add the "problem" explicitly to the test library.

Thanks.

pcanal:
The latest code here fails (for me, on the jenkins node) without https://github.com/root-project/root/pull/12863

hahnjo (approved):

PR #976, opened 2023-06-26
[RF] Make `roottest` again compile in memory-safe RooFit mode

guitargeek:
Wrap return values of `RooAbsPdf::generate()` directly in unique pointers.


pcanal (approved):

guitargeek:
Thanks!

PR #978, opened 2023-07-07
[skip-ci][DF] Remove unnecessary check for type of variable

vepadulano:
We only need to know that the event loop hasn't been triggered yet, the type of the proxy is an implementation detail.

### DIFF
--- a/python/distrdf/dask/check_reducer_merge.py
+++ b/python/distrdf/dask/check_reducer_merge.py
@@ -7,7 +7,6 @@ import numpy
 import ROOT
 
 from DistRDF.Backends import Dask
-from DistRDF.Proxy import ActionProxy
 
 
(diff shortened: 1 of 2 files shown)


eguiraud (approved):
LGTM.

As a note, it's not necessarily bad/weird to assert on implementation details -- I mean, if `npy_lazy` were _not_ an ActionProxy something evidently went wrong,so the assert is doing a little bit of useful work...so I'm not sure I understand the rationale for the change (e.g. are you planning to break that assert soon? if so the commit message could say so).

EDIT:
IIUC the rationale is indeed a naming change, https://github.com/root-project/root/pull/13203 . No reason to withhold this information. But I also now noticed this PR is against roottest, not root, so I care a bit less :smile:

PR #983, opened 2023-07-20
[ci] Fix specification of nbconvert

vepadulano:
Use the `!=` syntax to exclude the unwanted version of `nbconvert` in requirements.txt since there is no way to specify a logical OR operator in pip requirements.


hahnjo (approved):
Yes, please merge.

PR #984, opened 2023-07-20
Fix nbconvert version specifier in requirements.txt

hahnjo:
A comma is interpreted as a logical AND and naturally there exists no version that satisfies the requirement `< 7.3 , >= 7.4`. Change it to exclude `7.3.*`.


vepadulano:
Thank you Jonas! See #983 , I agree this it the right fix, we can merge either PR

PR #991, opened 2023-07-28
[cmake] Standardize CMake code

bellenot:
Remove special cases for Windows and make it build and run in both builtin and standalone mode


CMakeLists.txt
```diff
@@ -56,11 +56,7 @@ if(MSVC)
     # ROOT_BINDIR is set by ROOTConfig.cmake
     get_filename_component(realfp "${ROOT_BINDIR}/../" ABSOLUTE)
     file(TO_CMAKE_PATH ${realfp} ROOTSYS)
-    # FIXME: Get from root-config.bat --tutdir
```
hahnjo, line 59: Don't we want to keep this as a reminder?

bellenot: I tried, but `root-config.bat --tutdir` returns just `tutorials` in buit-in mode on Windows...

hahnjo: But shouldn't that be fixed instead? Does it work with the stand-alone build?

bellenot: Yes, it works in stand-alone. I'll check what's different between Linux and Windows...

bellenot: But it shouldn't really matter anyway (it's  semantically the same)

bellenot: @hahnjo I'm fixing `root-config.bat.in` right now. Will fix roottest afterwards

bellenot:
closing this PR, will make a new one when `root-config.bat` is fixed and verified that everything is fine with both builtin and standalone roottest

PR #994, opened 2023-07-31
[PyROOT] Compare floating point numbers from numba with isclose

hahnjo:
This will fix the test on macOS with LLVM 16 where interpreted code makes use of fma instructions that yield slightly different results.


sudo-panda (approved):
LGTM!

PR #995, opened 2023-07-31
[DF] Update test with behaviour of new unified constructor

vepadulano:
With ROOT7 enabled, The RDataFrame constructor taking (std::string_view
treename, std::string_view filename) now opens the file to check whether
the dataset is a TTree or an RNTuple. The warnings about the missing
dictionaries will now happen at construction time.

Create a new .ref file and instruct cmake to use the right one depending
on the ROOT configuration.

Sibling PR of https://github.com/root-project/root/pull/13090


dpiparo (approved):
LGTM.

PR #996, opened 2023-08-03
[RF] Adapt to changes in BatchMode

guitargeek:
Change the code so that it compiles also after merging the following PR in `root`:
https://github.com/root-project/root/pull/13389


lmoneta (approved):
Lgtm

PR #999, opened 2023-08-16
[skip-ci] dataframe: Disable ACLiC rootmap generation on Windows

hahnjo:
for https://github.com/root-project/root/pull/13460

### DIFF
--- a/root/dataframe/.rootrc
+++ b/root/dataframe/.rootrc
@@ -1,1 +1,2 @@
 Rint.History:            .root_hist
+ACLiC.LinkLibs: 1


root/dataframe/CMakeLists.txt
```diff
@@ -230,6 +230,6 @@ ROOTTEST_ADD_TEST(test_progressiveCSV
 
 # need the '+' to autogenerate required dictionaries with ACLiC
 ROOTTEST_ADD_TEST(test_nested_rvec_snapshot MACRO test_nested_rvec_snapshot.C+)
-ROOTTEST_ADD_TEST(test_snapshot_copyaddresses MACRO test_snapshot_copyaddresses.C+)
+ROOTTEST_ADD_TEST(test_snapshot_copyaddresses MACRO test_snapshot_copyaddresses.C)
```
eguiraud, line 233: this is going to confuse people (e.g. me) in the future. `test_nested_rvec_snapshot.C` needs ACLiC so it's "allowed" to use it, but for most other tests ACLiC is disallowed "because Windows"? :thinking: what's the rule? or is this temporary and a Windows fix is coming?

hahnjo: The rule seems to be that no two ACLiC compiled macros must define the classes or variables of the same name - but I need to confirm that it actually works on Windows. For this change I just disabled ACLiC on all tests and then added back the ones that really need it due to dictionaries.

eguiraud: > no two ACLiC compiled macros must define the classes or variables of the same name

two ACLiC compiled macros in the same directory, that is, right? sounds like a bug, can we point to the github issue from the commit message to keep track of things? (again thinking of future devs getting confused about this)

pcanal: This is not the best solution (neither rename nor disabling ACliC).   Instead, one "just" need to disable rootmap file for ACliC by adding:
```
ACLiC.LinkLibs: 1
```
to the `.rootrc` in that directory.

hahnjo: Aha, the default is `ACLiC.LinkLibs: 3` only on Windows... Ok, let's try that.

hahnjo:
https://github.com/root-project/root/pull/13460 looks good now, can somebody approve this PR?

bellenot (approved):
Thanks Jonas!

PR #1004, opened 2023-08-22
[Meta] Fix test on Windows by using the correct, platform independent way, of referring to the root command

dpiparo:
On this platform the copying of files does not work. For this reason, the test is disabled in order to keep the CI working in a healthy way.

### DIFF
--- a/root/meta/rootmap/duplicateWarning/CMakeLists.txt
+++ b/root/meta/rootmap/duplicateWarning/CMakeLists.txt
@@ -2,5 +2,5 @@ configure_file(rfile1.rootmap . COPYONLY)
 configure_file(rfile2.rootmap . COPYONLY)
 
 ROOTTEST_ADD_TEST(myclassduplicated
-                  COMMAND root -b -q -l
+                  COMMAND ${ROOT_root_CMD} -b -q -l
                   PASSREGEX  "Warning in <TInterpreter::ReadRootmapFile>: While processing ./rfile[12]\.rootmap, class myclassduplicated was found to be associated to libdup[12]\.so although it is already associated to libdup[12]\.so")


hahnjo (changes requested):
What error message says that copying files doesn't work? https://lcgapp-services.cern.ch/root-jenkins/job/roottest-pullrequests-build/13001/testReport/projectroot.roottest.root.meta.rootmap/duplicateWarning/roottest_root_meta_rootmap_duplicateWarning_myclassduplicated/ says that `C:/build/workspace/roottest-pullrequests-build/build/roottest/root/meta/rootmap/duplicateWarning/root` could not be found, which is entirely correct.

As I wrote on Mattermost, I think it's just a matter of using `root.exe` or better `${ROOT_root_CMD}`

hahnjo (approved):
Thanks!

root/meta/rootmap/duplicateWarning/CMakeLists.txt
```diff
@@ -2,5 +2,5 @@ configure_file(rfile1.rootmap . COPYONLY)
 configure_file(rfile2.rootmap . COPYONLY)
 
 ROOTTEST_ADD_TEST(myclassduplicated
-                  COMMAND root -b -q -l
+                  COMMAND ${ROOT_root_CMD} -b -q -l
                   PASSREGEX  "Warning in <TInterpreter::ReadRootmapFile>: While processing ./rfile[12]\.rootmap, class myclassduplicated was found to be associated to libdup[12]\.so although it is already associated to libdup[12]\.so")
```
bellenot, line 6: Not sure about the `.so` suffix on Windows. It will most probably complain about `.dll`

bellenot: Well, in fact it doesn't

bellenot (approved):

root/meta/rootmap/duplicateWarning/CMakeLists.txt
hahnjo: I think it's because `root/meta/rootmap/duplicateWarning/rfile1.rootmap` and `rfile2.rootmap` literally specify `libdup1.so` and `libdup2.so`, respectively.

bellenot: Yes, I saw it (a bit too late unfortunately, hence the comment and then solving the conversation. Sorry for the trouble

PR #1005, opened 2023-08-24
[PyROOT][9058] Add a test for TPython::ExecScript

dpiparo:


dpiparo:
This refers to the ROOT PR https://github.com/root-project/root/pull/13532

python/tpython/execscript.py
```diff
@@ -0,0 +1,8 @@
+import sys
+argc = len(sys.argv)
+
+if len(sys.argv) == 3 and sys.argv[1] == "foo" and sys.argv[2] == "bar":
+    print("SUCCESS: the arguments were foo and bar.")
+else:
+    print("FAILURE: the arguments were %s, while foo and bar were expected." %sys.argv)
+    sys.exit(1)
```
vepadulano, line 8:
```suggestion
import os
import sys


def main():
    filename = os.path.basename(__file__)
    if sys.argv != [filename, "foo", "bar"]:
        raise ValueError(
            f"Invalid arguments passed. This script must be called as '{sys.executable} {filename} foo bar'")


if __name__ == "__main__":
    raise SystemExit(main())

```

vepadulano (approved):
Thanks Danilo! See suggestion on improvement for the Python script

PR #1008, opened 2023-08-25
[RF] Don't exclude vectorized pdf tests on Windows

guitargeek:
There should be no reason anymore for these tests not running on Windows.


bellenot (approved):

guitargeek:
@phsft-bot build

guitargeek:
Let's retry, now that the tests were updated to use much less events (maybe that makes a difference)

PR #1011, opened 2023-09-04
[PyROOT] Adapt tests to the new default memory policy "strict"

guitargeek:
Sister PR of https://github.com/root-project/root/pull/13593.


guitargeek:
@phsft-bot build

guitargeek:
Not needed anymore because we don't consider to change the memory policy right now.

PR #1021, opened 2023-09-29
Use non-deprecated (and now removed) version of failUnlessEqual:

Axel-Naumann:
See https://docs.python.org/2/library/unittest.html#deprecated-aliases

### DIFF
--- a/python/numba/CMakeLists.txt
+++ b/python/numba/CMakeLists.txt
@@ -1,8 +1,10 @@
 if(ROOT_pyroot_FOUND)
-  if(NOT MSVC OR win_broken_tests)
-    ROOTTEST_ADD_TEST(numba
-                      MACRO PyROOT_numbatests.py
-                      ENVIRONMENT LEGACY_PYROOT=${legacy_pyroot})
-    file(COPY vec_lv.root DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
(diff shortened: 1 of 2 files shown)


python/numba/CMakeLists.txt
```diff
@@ -1,8 +1,10 @@
 if(ROOT_pyroot_FOUND)
-  if(NOT MSVC OR win_broken_tests)
-    ROOTTEST_ADD_TEST(numba
-                      MACRO PyROOT_numbatests.py
-                      ENVIRONMENT LEGACY_PYROOT=${legacy_pyroot})
-    file(COPY vec_lv.root DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+  if(NOT (DEFINED ENV{ROOTTEST_IGNORE_NUMBA_PY3} AND PYTHON_VERSION_MAJOR_Development_Main EQUAL 3))
```
vepadulano, line 2: This variable is fragile, as it is only defined in https://github.com/root-project/rootspi/blob/c4b603d131efdd97f2b2008250adbbd39e86d5a8/jenkins/root-build.cmake#L300 . For it to make sense also in the new CI builds we would need to export the environment variable there too. In general, we need to rethink this system probably

vepadulano (approved):
LGTM thanks! But see comment

python/numba/CMakeLists.txt
Axel-Naumann: https://github.com/root-project/root-ci-images/blob/main/fedora39/Dockerfile#L24

:-)

PR #1026, opened 2023-10-06
[6.28] Backport of 1025

vepadulano:
The distributed RDataFrame test batteries use pytest to run all the tests together as a single ctest test. This is crucial to ensure only one mock cluster connection is open for all the needed tests. Sometimes, these test batteries may take much longer than expected. For example, on older and less powerful machines such as the ones running MacOS 11, running one distributed test battery usually takes 3 minutes, while it may take up to 10 minutes in case the node is overbooked and different process happen at the same time. Thus, this commit increases the timeout threshold to 20 minutes, in line with some of the longest running tests in our CI suite. This should ensure that scenarios with a very slow machine but no real test failure still work.


dpiparo (approved):
Thanks Vincenzo. This is needed by our CI.

PR #1027, opened 2023-10-06
[Cppyy][13410] Test accessibility of derived class

vepadulano:
Sibling of https://github.com/root-project/root/pull/13821 with test taken from https://github.com/root-project/roottest/pull/1024


vepadulano:
#1024 was merged

PR #1043, opened 2024-01-12
Update `tree-fastcloning-execCheckClusterRange` reference files

guitargeek:
In `root-project/root`, we are changing some of the TLeaf data member doc strings.

These class comments are part of the TStreamerInfo, which can affect compressed size as explaine by Philippe Canal in this comment: https://github.com/root-project/root/pull/14268#issuecomment-1865251700

Sister PR of [#14268](https://github.com/root-project/root/pull/14268).


pcanal (approved):
Good to go once the test both here and the corresponding root-project/root PR passes all the tests.

hahnjo:
This is missing an update of the win64 reference files; now done in https://github.com/root-project/roottest/pull/1044

pcanal:
Thanks.

PR #1046, opened 2024-01-19
[DF] Split test_snapshot into multiple tests to avoid timeout.

martamaja10:
It was noticed that test_snapshot was quite often timeout'ed during the CI. The idea now is to split this test into 4 smaller ones to avoid the issue.


root/dataframe/test_snapshotMT_dir.C
```diff
@@ -0,0 +1,154 @@
...
+
+  return 0;
+}
+
+int runTest() {
+  auto fileName = "test_snapshot.root";
+  auto outFileName = "test_snapshot_output.root";
+  auto treeName = "myTree";
+  auto outTreeName = "myTree";
+  fill_tree(fileName, treeName);
```
vepadulano, line 132: In these tests, whenever we fill a file it would be nice to have the guarantee that the file is destroyed at the end of the test, for safety and tidiness. This can be done with a RAII structure such as the one found here

https://github.com/root-project/root/blob/7c702f8302a9da73ca694d1413f828b27d5e5fdd/tree/ntuple/v7/test/ntuple_test.hxx#L116

And then used throughout tests like https://github.com/root-project/root/blob/7c702f8302a9da73ca694d1413f828b27d5e5fdd/io/io/test/RRawFile.cxx#L46


In principle the structure could be augmented by also putting the code of `fill_tree` inside it e.g. in the constructor of the class.

root/dataframe/test_snapshotMT_dir.C
```diff
@@ -0,0 +1,154 @@
...
+
+  std::cout << "---- Now with a tree in the root directory\n";
+  int ret = do_work(fileName, outFileName, treeName, outTreeName);
+
+  // now we put the tree in a directory
+  outFileName = "test_snapshot_inDirectory_output.root";
+  treeName = "myTree";
+  outTreeName = "a/myTree";
+
```
vepadulano, line 141: I don't understand this part, the comment says that a tree is going to be put in a directory, but I don't see any code creating a tree or a file

vepadulano (changes requested):
A few starting comments

root/dataframe/test_snapshotMT_dir.C
martamaja10: my bad, this part should be deleted as it was moved to the separate subdir test

root/dataframe/dataframe_test.hxx
martamaja10, line 1: I got inspired by this one: https://github.com/root-project/root/blob/master/io/io/test/io_test.hxx and left the minimal needed functionality, but maybe something is missing here?

martamaja10:
the new CI does not error on those tests: https://github.com/root-project/root/actions/runs/7585086531

root/dataframe/dataframe_test.hxx
```diff
@@ -0,0 +1,23 @@
+#include <cstring>
```
vepadulano, line 1: I acknowledge this header is only going to be included by these tests. I would still place header guards at beggining and end, it's just a good practice (that the original header you link to doesn't follow, I also acknowledge that)

vepadulano, line 1: Is the `cstring` header needed?

vepadulano (approved):
Thank you @martamaja10 ! Just a couple of small suggestions left

PR #1048, opened 2024-02-01
[cmake] Skip its types if Proof is disabled

dpiparo:


hahnjo (approved):
@dpiparo I think we need this together with https://github.com/root-project/root/pull/14513 - I'm not sure if it solves all test failures though, `roottest-root-io-compression-make` fails because `root/io/compression/MainEvent.cxx` includes `TProofServ.h` (not clear why) and `roottest-root-tree-friend-make` due to `root/tree/friend/testFriendsIndices.C` including `TDSet.h` (but the actual test code is commented out)

dpiparo:
sorry. Yes.

PR #1051, opened 2024-02-05
[PyROOT] New unit test with reproducer from ROOT issue 14232

guitargeek:
See https://github.com/root-project/root/issues/14232.


guitargeek:
We should probably merge this PR, since there is no failure in the CI, and like this we can maybe see the problem in the nightlies at some point (or just close the linked issue if the problem with the `tuple_element` turns out to not be reproducible at all).

dpiparo:
I agree.

guitargeek:
@dpiparo, can you please approve this PR so I can merge it? Thanks

dpiparo (approved):
LGTM! thanks for this nice test.

PR #1052, opened 2024-02-05
[PyROOT] Add reproducer for ROOT issue 11411 as a unit test

guitargeek:
See https://github.com/root-project/root/issues/11411


guitargeek:
The CI clearly fails when adding the reproducer from the issue, therefore the issue is confirmed.

PR #1061, opened 2024-02-14
[PyROOT] Don't use the `TDirectory.__getattr__` pythonization

guitargeek:
This allows us to later deprecate it in the main ROOT repo without requiring another PR in roottest.

Also, explicit tests for the `TDirectory.__getattr__` are removed, because this is already covered by pythonization tests in the main repo:

  * https://github.com/root-project/root/blob/master/bindings/pyroot/pythonizations/test/tdirectory_attrsyntax.py

  * https://github.com/root-project/root/blob/master/bindings/pyroot/pythonizations/test/tdirectoryfile_attrsyntax_get.py

### DIFF
--- a/python/numba/PyROOT_numbatests.py
+++ b/python/numba/PyROOT_numbatests.py
@@ -61,7 +61,7 @@ class TestClasNumba:
         # Obtain a vector of ROOT::Math::LorentzVector from the sample
         # .root file
         myfile = ROOT.TFile.Open("vec_lv.root")
-        vec_lv = myfile.vecOfLV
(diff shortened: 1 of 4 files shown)


guitargeek:
The test failures are unrelated to this PR.

There are from the associated main root PR: https://github.com/root-project/root/pull/14660

Meaning this PR can be merged safely.

guitargeek:
Re-opening with a different branch name, so it doesn't get tested together with the draft PR in the main ROOT repo:
https://github.com/root-project/roottest/pull/1073

PR #1064, opened 2024-02-19
Remove THnL64

ferdymercury:
as suggested by lmoneta. THnL will now be directly based on TArrayL64


ferdymercury:
This is needed for https://github.com/root-project/root/pull/8546

PR #1068, opened 2024-02-22
Try to fix build failure on Windows

bellenot:


pcanal (approved):
No quite clear why it fails on Windows.  But this looks clearer and more explicit.

PR #1071, opened 2024-02-29
[PyROOT] Update cppyy to a recent version

guitargeek:
Sister PR of https://github.com/root-project/root/pull/14507.

### DIFF
--- a/python/cpp/PyROOT_cpp11tests.py
+++ b/python/cpp/PyROOT_cpp11tests.py
@@ -58,6 +58,18 @@ class Cpp1Cpp11StandardClassesTestCase( MyTestCase ):
 
         ROOT.std.tuple_element[1, ATuple].type
 
+    def test03InitializerList(self):
+        """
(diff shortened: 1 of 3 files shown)


python/cpp/PyROOT_cpp11tests.py
```diff
@@ -58,6 +58,18 @@ def test02TupleElement(self):
 
         ROOT.std.tuple_element[1, ATuple].type
 
+    def test03InitializerList(self):
+        """
+        Check if we can convert a list of strings to a Python initializer list.
```
vepadulano, line 63: C++ initializer list

vepadulano (approved):
Approving as the sibling PR was approved

python/cpp/PyROOT_cpp11tests.py
guitargeek: Ah yes, you're right.

PR #1073, opened 2024-03-01
[PyROOT] Don't use the `TDirectory.__getattr__` pythonization

guitargeek:
This allows us to later deprecate it in the main ROOT repo without requiring another PR in roottest.

Also, explicit tests for the `TDirectory.__getattr__` are removed, because this is already covered by pythonization tests in the main repo:

  * https://github.com/root-project/root/blob/master/bindings/pyroot/pythonizations/test/tdirectory_attrsyntax.py

  * https://github.com/root-project/root/blob/master/bindings/pyroot/pythonizations/test/tdirectoryfile_attrsyntax_get.py


guitargeek:
Test failures are unrelated.

vepadulano (approved):
LGTM thanks!

PR #1076, opened 2024-03-08
[RF] Reduce numer of events in the vectorised pdf tests

guitargeek:
Reduce numer of events in the vectorised pdf tests, and accordingly increase the post-fit tolerance a bit, because the statistical fluctuations are higher.

This cuts down the total time of the tests on one CI node by about 14 minutes.

Furthermore, the vectorized evaluation is now validated so much - also in the full integration test `stressRooFit` - that these tests are mostly redundant.


lmoneta (approved):
LGTM!
It does not make much sense to generate so many events, it is good you reduce it to shorten the CI time!

PR #1086, opened 2024-03-14
add test for thnbase printentries

ferdymercury:
matches https://github.com/root-project/root/pull/14957


dpiparo:
Thanks a lot @ferdymercury !

lmoneta (approved):

PR #1090, opened 2024-03-18
[tree] test for limits in buildindex

ferdymercury:


root/tree/index/runindex64.C
```diff
@@ -8,9 +8,11 @@ const char* fname = "index64.root";
 // Apple M1 has long double == double; these values exceed its range
 // and cannot be represented as (even temporary) expression results.
 // There would be a warning if you'd try.
-static constexpr bool shortlongdouble = sizeof(long double) < 16; // was true for __APPLE__ and __arm64__
-const Long64_t bigval   = shortlongdouble ? 0xFFFFFFFFFFFF : 0xFFFFFFFFFFFFFFF; // still positive number
+static constexpr bool shortlongdouble = sizeof(long double) < 16 || true; // was true for __APPLE__ and __arm64__
```
vepadulano, line 11: Why the `|| true` ?

root/tree/index/runindex64.C
```diff
@@ -8,9 +8,11 @@ const char* fname = "index64.root";
...
+const Long64_t bigval   = shortlongdouble ?  0xFFFFFFFFFFFF :  0xFFFFFFFFFFFFFFF; // still positive number
```
vepadulano, line 12: What is this condition? How does `bigval` affect the reading and under which circumstances?

root/tree/index/runindex64.C
```diff
@@ -8,9 +8,11 @@ const char* fname = "index64.root";
...
 const ULong64_t biguval = shortlongdouble ?  0xFFFFFFFFFFFF0 : 0xFFFFFFFFFFFFFFF0; // "negative" number
+const Long64_t maxMajor = 0x1ffffffff;
+const Long64_t maxMinor = 0x7fffffff; // unless major = 0
```
vepadulano, line 15: If `major=0`, what happens to `maxMinor`? It's unclear

root/tree/index/runindex64.C
```diff
@@ -25,20 +27,22 @@ int runindex64(){
   tree->Branch("run", &run, "run/l");
   tree->Branch("event", &event, "event/l");
 
-  ULong64_t events[] = { 1,2,3, bigval, biguval, 5 };
-  run = 5;
+  ULong64_t   runs[] = { 8,5,5,5,      0,       0,  5,      4, 6, biguval,          1, maxMajor, maxMajor+1};
+  ULong64_t events[] = { 0,1,3,2, bigval, biguval,  5, bigval, 3,  bigval, maxMinor+1, maxMinor, 0};
```
vepadulano, line 31: This is a weird layout, did this go through `clang-format`?

vepadulano, line 31: These should become `std::array<ULong64_t, 13>`

root/tree/index/runindex64.C
```diff
@@ -25,20 +27,22 @@ int runindex64(){
...
   for(int i=0; i<sizeof(events)/sizeof(*events); i++){
+    run = runs[i];
     event = events[i];
     tree->Fill();
```
vepadulano, line 35: With `events` being an `std::array` this simply becomes
```cpp
std::for_each(
    std::begin(events),
    std::end(events),
    [i=0] (ULong64_t val) mutable { run = runs[i]; event = val; tree->Fill(); i++;}
)
```

root/tree/index/runindex64.C
```diff
@@ -25,20 +27,22 @@ int runindex64(){
...
+  //tree->Scan("run:event","","colsize=30");
 
   cout<<"Tree BuildIndex returns "<<tree->BuildIndex("run", "event")<<endl;
-  cout << "Entry should be 3: " << tree->GetEntryNumberWithIndex(5,bigval) << endl;
-  cout << "Entry should be 6: " << tree->GetEntryNumberWithIndex(4,bigval) << endl;
+  for (int i=0; i<sizeof(events)/sizeof(*events); i++) {
+    run = runs[i];
+    event = events[i];
+    cout << i << ": Run " << run << ", Event " << event << " found at entry number: " << tree->GetEntryNumberWithIndex(run, event) << endl;
+  }
```
vepadulano, line 45: With `events` being an `std::array` this simply becomes
```cpp
std::for_each(
    std::begin(events),
    std::end(events),
    [i=0] (ULong64_t val) mutable {
        std::cout << i << ": Run " << runs[i] << ", Event " << val << " found at entry number: " << tree->GetEntryNumberWithIndex(run, event) << std::endl;
        i++;
    }
)
```

root/tree/index/runindex64.C
```diff
@@ -62,6 +66,7 @@ bool test(TTree *chain)
   cout<<"BuildIndex returns "<<chain->BuildIndex("run", "event")<<endl;
   cout<<"Try to get value that is not in the chain, this should return a -1:"<<endl;
   cout<<chain->GetEntryWithIndex(500)<<endl;
-  cout<<(int)chain->GetEntryNumberWithIndex(5, bigval)<<endl;
+  cout<<"Try to get value that is in the chain, this should return a 4:"<<endl;
```
vepadulano, line 69: I do not understand the meaning of this sentence sorry

root/tree/index/runindex64.C
```diff
@@ -62,6 +66,7 @@ bool test(TTree *chain)
...
+  cout<<(int)chain->GetEntryNumberWithIndex(0, bigval)<<endl;
```
vepadulano, line 70:
```suggestion
  cout<<chain->GetEntryNumberWithIndex(0, bigval)<<endl;
```

vepadulano (dismissed):
Thanks! Some questions regarding the implementation of the test

root/tree/index/runindex64.C
ferdymercury: because otherwise, the roottest fails, as it's platform dependent, and there is only one index64.ref to compare with.

root/tree/index/runindex64.C
ferdymercury: This was here before. Apparently, TTree::Print() goes into overflow at different max-values, depending on the platform.

root/tree/index/runindex64.C
pcanal: This change the nature of the next few lines and essentially remove their use and what they were testing.  For example see 9e3843d4bf50bc34e6e15dfe7c027f029417d6c0.

The length of long double is platform dependent and thus indeed the result/behavior might be (the next 2 lines were the way to deal with it).  We may have to switch from relying on the output to more direct comparison for the new code to handle properly the platform dependence.

root/tree/index/runindex64.C
pcanal: See f3c97809c9064feccaed3844007de9e7c6a5980d for some explanation.

root/tree/index/runindex64.C
pcanal: That's actually a nice layout here where you can see the pair of values that goes together without having to count commas.

root/tree/index/runindex64.C
pcanal: On the other hand for this trivial example, it is not clear that the modern C++ is more 'readable' that the existing code.  More over, this change is not directly related to the goal of the PR.

root/tree/index/runindex64.C
ferdymercury: thanks! But one advantage of this formulation over std::array is that you don't have to count nr of elements (13), it's just auto-initialized. see https://comp.lang.cpp.moderated.narkive.com/ZMoUm1AY/std-array-if-only-it-knew-its-size

pcanal (changes requested):
See comments.

root/tree/index/runindex64.C
ferdymercury: What do you think about adding a switch here: https://github.com/ferdymercury/roottest/blob/majorminor/root/tree/index/Makefile
that calls 'sed' to replace the platform-differences in runIndex, if on Mac?

The other think I can think of is to define a custom ErrorHandler to equalize the things printed.

pcanal: > What do you think about adding a switch here .. Makefile.

I'll rather avoid increasing the complexity of the Makefile unless absolutely necessary.  In addition the `Makefile` are deprecated and will eventually be replaced by `CMake/CTest` fragment.

> The other think I can think of is to define a custom ErrorHandler to equalize the things printed.

I am a bit confused.  What is the code that lead to the difference? What is the actual difference?

ferdymercury: Note that event 5 is not found on Linux.

pcanal (changes requested):
In previous iteration this PR was blocked because the test output was not platform independent enough.  Did that get resolved?

ferdymercury:
I didn't manage to find a way to solve it :s

pcanal:

pcanal:

ferdymercury:
Summary of the platform-dependency.

It's not only that the numbers are different, but also that some are not found in one case and are found in the other.

```
-4: Run 5, Event 1152921504606846975 found at entry number: 4
+4: Run 5, Event 281474976710655 found at entry number: 4

-5: Run 5, Event 18446744073709551600 found at entry number: -1
+5: Run 5, Event 4503599627370480 found at entry number: 5

-7: Run 4, Event 1152921504606846975 found at entry number: 7
+7: Run 4, Event 281474976710655 found at entry number: 7

-9: Run 18446744073709551600, Event 1152921504606846975 found at entry number: -1
+9: Run 4503599627370480, Event 281474976710655 found at entry number: 9

-10: Run 1152921504606846975, Event 18446744073709551600 found at entry number: -1
+10: Run 281474976710655, Event 4503599627370480 found at entry number: 10

-11: Run 1152921504606846975, Event 1152921504606846975 found at entry number: 11
+11: Run 281474976710655, Event 281474976710655 found at entry number: 11

-12: Run 18446744073709551600, Event 18446744073709551600 found at entry number: -1
+12: Run 4503599627370480, Event 4503599627370480 found at entry number: 12
```

Is there a way in the run.C mechanism to express platform dependency with different logs? something like run$PLATFORM or so? I am out of ideas here.

vepadulano:
> Is there a way in the run.C mechanism to express platform dependency with different logs? something like run$PLATFORM or so?

I believe the best way is to define different tests depending on the platform in the CMakeLists.txt. I see that unfortunately this folder still uses the old Makefile way, so @pcanal is maybe able to guide us better in this sense.

One question though

> but also that some are not found in one case and are found in the other.

If finding an entry in the index is platform-dependent, isn't that a problem? e.g. Mac users might be running an analysis and getting different results than Linux users?

root/tree/index/runindex64.C
```diff
@@ -25,20 +26,22 @@ int runindex64(){
...
+  for(size_t i=0; i<sizeof(events)/sizeof(*events); i++){
+    run = runs[i];
     event = events[i];
     tree->Fill();
   }
-  run = 4; event = bigval; tree->Fill();
-  run = 6; event = 3; tree->Fill();
-  run = biguval; event = bigval; tree->Fill();
   tree->Write();
+  tree->Scan("run:event","","colsize=30");
```
pcanal, line 37: What is the intent of this addition? (I.e. what is it testing?)

ferdymercury: It shows that Scan works well, but somehow the entry is not foundnin Mac, even if the Scan part works well

pcanal:
> If finding an entry in the index is platform-dependent, isn't that a problem?

Indeed, the entry should be the same or at least the one intended otherwise there is a problem.

> Is there a way in the run.C mechanism to express platform dependency with different logs? 

There is several ways.  It easily supports 32bits vs 64bits and interpreted vs compiled but not really this vs that platform.  There is way to filter/modify the log before comparison.

However, the usual way to deal with the platform difference is to test the values directly within the code rather than indirectly via printing (the later works well when comparing a large number of values that are not platform dependent).

> -4: Run 5, Event 1152921504606846975 found at entry number: 4
> +4: Run 5, Event 281474976710655 found at entry number: 4

Here the difference is indeed spurrious as the intended entry number is found.  We could either not print the Run Number/Event Number (as the existing code did) or only print them if the entry number is not 4.

-5: Run 5, Event 18446744073709551600 found at entry number: -1
+5: Run 5, Event 4503599627370480 found at entry number: 5

The difference here seems to indicate an actual error (either in the test or in the code). (The reference file in the master seems to indicates that this used to work)

root/tree/index/runindex64.C
pcanal: Good, I guess.  This means this is a debugging tool and is not inherent to the test.  To keep the output platform independent we could run the `Scan` only in the cases where a failure is detected.

ferdymercury:
> (The reference file in the master seems to indicates that this used to work)

The reference file was casting to 32bit a number that was longer than 32bits so it was just working by chance, I believe.

ferdymercury:
> We could either not print the Run Number/Event Number

Yep I can do that. But first I wanted to understand the issue with Event5 not being found.

ferdymercury:
moved to main repo

PR #1093, opened 2024-03-26
[df] Homogenize distributed testing for different parameters

vepadulano:
Create a single set of testing scripts to run according to different parameters for distributed execution.

The biggest change with this commit is moving from two separate sets of files for Spark and Dask backends to a single set that can run with either backend connection depending on the current parameter in the `payload` fixture set via the `conftest.py` file.

Also, create a directory with input TTree datasets used throughout the tests. This has two advantages. First, it avoids spending extra time in creating and writing the dataset to disk. Second, it sets the ground for a future update of the testing suite to include the data source as another parameter (i.e. running the tests reading either TTree or RNTuple data).

For reproducibility, the `_create_datasets.py` script is added to the input data directory.


vepadulano:
All builds are green at the sibling PR, emerging this

PR #1095, opened 2024-03-28
[df] Remove unnecessary distributed test

vepadulano:
The `common` folder tests for a very specific, admittedly extremely rare or inexistent, use case. That is, an application that sends multiple tasks concurrently to different resource schedulers at the same time. Currently, the test submits the same RDataFrame graph to both a Spark scheduler and a Dask scheduler concurrently. So far, this scenario has never occurred in practice and perhaps the test is overzealous. Furthermore, it sporadically fails on the CI due to timeouts that do not provide any evidence of the root cause. For lack of concrete necessity, we can disable the test and eventually re-enable it at a later stage if required.


dpiparo (approved):
LGTM. Thanks for the throrough thinking about these tests.

PR #1096, opened 2024-03-28
[df] Test multiple tasks with AsNumpyResult

vepadulano:
Sibling PR of https://github.com/root-project/root/pull/15088


dpiparo (approved):
Thanks for this test, LGTM.

PR #1110, opened 2024-04-16
Fix failing execROOT_6277 on Windows

bellenot:
Compile the macro beforehand to prevent having to run the test twice,
the first time being always failing with the following error:
`LINK : fatal error LNK1561: entry point must be defined`
and successful the second time


bellenot:
@phsft-bot build just on windows10/default

bellenot:
> Thanks for this PR. Are we sure the dependency between the compilation and running the test is honoured?

After looking at the other tests and after running several times I would say yes :-)

dpiparo (approved):

PR #1111, opened 2024-04-21
Add test for ROOT-10425

dpiparo:
And, as per agreement, consider the item (https://its.cern.ch/jira/browse/ROOT-10425) fixed.


root/rint/CMakeLists.txt
```diff
@@ -12,3 +12,8 @@ if(${CMAKE_SYSTEM_PROCESSOR} MATCHES "x86_64.*|x86.*|amd64.*|AMD64.*|i686.*|i386
                     COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/driveTabCom.py
                     INPUT BackslashNewline_input.txt)
 endif()
+
+# ROOT-10425
+ROOTTEST_ADD_TEST(missingSymbol
+                 MACRO missingSymbol.C
+                 PASSREGEX "Maybe you need to load the corresponding shared library")
```
pcanal, line 19: This is likely sufficient but technically we should also check that the output is talking about the symbol we are intentionally missing.

dpiparo:
good suggestion. I changed the pass regex: it should work on the 3 platforms.

PR #1119, opened 2024-05-06
[CMake] Replace deprecated `FindPythonInterp` module

guitargeek:
See:
https://cmake.org/cmake/help/latest/module/FindPythonInterp.html

Sister PR of https://github.com/root-project/root/pull/15420.


guitargeek:
Forgot to use batching branch name, reopening: https://github.com/root-project/roottest/pull/1120

PR #1125, opened 2024-05-11
Revert "[PyROOT] Update expected exception in `TPyException` regressi…

guitargeek:
…on test"

This reverts commit 26a6ee9ecca7abf3ab16eefdeb04ab045c17b720, which is not necessary anymore since the change of exception types was fixed upstream in CPyCppyy.

Goes together with:
https://github.com/root-project/root/pull/15485


vepadulano (approved):
Nice!

PR #1128, opened 2024-05-30
[ntuple] add initial testing infrastructure

jblomer:
Add the ntuple subfolder with a first, trivial test. Open the scope for RNTuple integration tests.


hahnjo (approved):

pcanal (approved):

root/ntuple/test_basics.ref
```diff
@@ -0,0 +1,5 @@
+Processing /home/jakob/Documents/CERN/ROOT/roottest/root/ntuple/basics.C...
```
enirolf, line 1: Shouldn't this path be changed?

PR #1135, opened 2024-06-13
[cmake][win] Disable problematic tests

bellenot:
Disable a couple of tests that are triggering full (sort of) rebuild of ROOT


bellenot:
@dpiparo as discussed, to be merged if the problem of rebuilding LLVM/ROOT while compiling these tests is still an issue in the CI

PR #1142, opened 2024-07-09
[skip-ci] Add test for hadd input format validation

ferdymercury:
sibling of https://github.com/root-project/root/pull/16001


ferdymercury:
I am not sure on what structure this test needs to have, as I always did tests before with C++. I guess the output of the console should be somehow parsed and compared against a template, but I'll need help for this.

vepadulano:
Updated the test and I restarted the CI in the sibling ROOT PR

vepadulano (approved):

PR #1143, opened 2024-07-11
[ntuple] add test for changing RNTuple compression with hadd

silverweed:
Depends on #1140 and https://github.com/root-project/root/pull/16013
Adding a test for changing RNTuple compression with hadd; also restructure merge tests to use gtest (more reliable than comparing stdout).


jblomer (commented):
In principle looks good to me. I think I'd put the `.clang-format` file in a separate PR.

Following our discussion, do you plan to restructure the test into several ctest invocations?

silverweed:
@jblomer I agree on clang-format, I'll split it into another PR. 
For the second thing, I'll see what that implies and decide if that works well.

silverweed:
@jblomer clang-format PR: #1144

silverweed:
@jblomer ctest PR: #1145 1145

hahnjo:
I think this can be closed after https://github.com/root-project/roottest/pull/1145 was merged...

PR #1145, opened 2024-07-12
[ntuple] change merge tests to use PRECMD/POSTCMD

silverweed:
Followup of #1143 

For some reason it takes ~13s to JIT each macro on my machine (in release mode -- 20s in debug)


vepadulano:
Writing this comment here but it applies to all related PRs. IMHO it would be better to group together all `hadd` tests together in the `root/io/hadd` directory, even though this one in particular focuses on RNTuple.

root/io/hadd/CMakeLists.txt
```diff
@@ -1,6 +1,18 @@
...
 endif()
 
-ROOTTEST_ADD_TEST(compression_settings
-                  MACRO test_compression_settings.py)
\ No newline at end of file
+ROOTTEST_ADD_TEST(test_Merge
+                  COPY_TO_BUILDDIR merge_gen_input_tuples.C merge_check_output.C
+                  PRECMD ${ROOT_root_CMD} -q -b -l "merge_gen_input_tuples.C+(\"test_merge_in1.root\", \"test_merge_in2.root\")"
+                  COMMAND ${ROOT_hadd_CMD} -f test_merge_out.root test_merge_in1.root test_merge_in2.root
+                  POSTCMD ${ROOT_root_CMD} -q -b -l "merge_check_output.C+(404, \"test_merge_out.root\", \"test_merge_in1.root\", \"test_merge_in2.root\")"
```
jblomer, line 10: Why is the expected compression LZ4 if the inputs are ZSTS/ZLIB compressed?

root/io/hadd/merge_check_output.C
```diff
@@ -0,0 +1,82 @@
...
+}
+
+bool FloatAlmostEquals(float a, float b)
+{
+   uint32_t ia, ib;
+   memcpy(&ia, &a, sizeof(float));
+   memcpy(&ib, &b, sizeof(float));
+   return DistanceBetweenSignAndMagnitudeNumbers(ia, ib) <= kMaxUlps;
+}
+} // namespace
```
jblomer, line 47: Perhaps it's easier to write the tests without floats.

root/io/hadd/merge_check_output.C
```diff
@@ -0,0 +1,82 @@
...
+
+int merge_check_output(int expectedCompression, const char *fnameOut, const char *fnameIn1,
```
jblomer, line 49: The expected compression is unused.

jblomer (commented):
In principle looks good, some smaller comments.

Regarding ACLiC time: wouldn't it make sense to not compile the macros at all and run them interpreted?

root/io/hadd/merge_check_output.C
silverweed: Copy-paste error, thanks for catching it!

jblomer (approved):

PR #1148, opened 2024-07-23
[io] update test expectations for hadd after #16082

silverweed:
See https://github.com/root-project/root/pull/16082


hahnjo (approved):

silverweed:
@hahnjo this PR simply corrects a stdout mismatch in an existing test due to 16082 changing the output message; the other one adds a new test specific for the compression behavior change (so, to test the addressed bug is actually fixed)

hahnjo:
Ok, please note that our CI doesn't support that: It will look for the first branch of matching name that it finds. Given that https://github.com/root-project/root/pull/16082 was green, it probably took this one...

PR #1150, opened 2024-07-24
[hadd] add test for hadd compression algorithm change for TTree

silverweed:
see https://github.com/root-project/root/pull/16082


jblomer (approved):
In principle LGTM. Perhaps let's make the 101 initial compression explicit. Otherwise, when we change the default setting, the test case loses meaning.

dpiparo (approved):
LGTM

PR #1157, opened 2024-08-02
Full- chain testing with headless chrome browser

DuesselbergAdrian:
DO NOT MERGE!!! This PR adds testing of new SVGs on Mac. It includes:
- Reference files for new SVGs
- A new function test_new_svg() to create svgs from reference json files and then compare them.
- A manual script how create new reference files


linev:
@DuesselbergAdrian 

I am very close to make testing of JSROOT ready with headless browsers.
It looks promising - for the moment I just cleanup JSROOT code that produced SVG/PNG/JSON files match on browser and with node.js.

Here my working macro:

https://github.com/linev/jsroot-test/blob/puppeteer/puppeteer.js

Probably then one can do very similar with ROOT http server.

PR #1159, opened 2024-08-06
[meta] Update reference file for runMemberComments.C

aaronj0:
Part of the fix for
- https://github.com/root-project/root/issues/15919


pcanal:
This changes the output to print/expect inaccurate information.  See https://github.com/root-project/root/pull/16177 for more details

PR #1160, opened 2024-08-07
added missing include

cburgard:
this PR resolves https://github.com/root-project/root/issues/16182


guitargeek (approved):
Thanks!

  * adds missing include for `std::setprecision` used in this file

PR #1162, opened 2024-08-09
[PyROOT] Update error style in numba derived class call

aaronj0:
This is to keep the test error style backward compatible with both numba <=0.59 and >=0.6 and prevent failures when enabling numpy 2.0

Fixes https://github.com/root-project/root/issues/16201


guitargeek (approved):
Thanks! Good you figured this out.

PR #1164, opened 2024-08-09
[v632] [PyROOT] Update error style in numba derived class call

aaronj0:
This is to keep the test error style backward compatible with both numba <=0.59 and >=0.6 and prevent failures when enabling numpy 2.0


guitargeek:
@dpiparo, can this be merged today? Otherwise, we'll have red 6.32 nightlies, as we're updating numba in the docker images tonight:
https://github.com/root-project/root/pull/16209

PR #1166, opened 2024-08-15
[core] Add test for ROOT-3594

dpiparo:
thanks to the original author for the initial code.


pcanal (approved):

dpiparo:
this PR is a good test but if merged w/o the fix in ROOT, it will break all builds. So it's merging it's dependent on the destiny of the ROOT PR mentioned above.

PR #1169, opened 2024-08-16
Fix wrong selection rules

dpiparo:
which can generate warnings as unused and make the test fail in case the verbosity level of Aclic is increased.


hahnjo (approved):
I believe this is needed to fix `roottest-cling-dict-runalgebra` after merging https://github.com/root-project/root/pull/16208

PR #1173, opened 2024-08-20
[cmake] Make some tests run serial on Windows

bellenot:
Should decrease the number of retries when running the tests on Windows


dpiparo (approved):
lgtm, thanks.

PR #1174, opened 2024-08-21
Update roottest for the disabling of THtml

silverweed:
Update roottest for the disabling of THtml (will squash the commits)


guitargeek (approved):
Nice, this looks good! We should merge this soon to get green builds, but also try to validate that the modified test still covers the original issue.

PR #1175, opened 2024-08-23
Test for the .libraries command for the prompt

dpiparo:
test for ROOT-5843 (see ROOT PR https://github.com/root-project/root/pull/16301 )


pcanal (approved):
The test has the intended functionality.  It could be (later?) extended to not only test that `.libraries` exist but it also works properly in edge case (indirectly loaded libraries, libraries without dictionary, dlopen-ed libraries loaded directly or indirectly, working on Windows, etc.)

dpiparo:
absolutely. Actually we could test all the `.command`

PR #1184, opened 2024-09-10
[CMake] Set the minimum required version to 3.20

dpiparo:
This is required to build LLVM18.
This version of CMake was eleased in March 2021 https://github.com/Kitware/CMake/releases/tag/v3.20.0

Associated PR in ROOT https://github.com/root-project/root/pull/16396


hahnjo:
we may need some `cmake_policy` magic here, some defaults are changed based on `cmake_minimum_required`...

PR #1190, opened 2024-09-24
disable some geom-dependant tests when geom is OFF

silverweed:
Adapt roottests to geom being off. Related to [#16514](https://github.com/root-project/root/pull/16514)


dpiparo (approved):
LGTM, Thanks.

PR #1200, opened 2024-10-14
[skip-ci] [df] Adapt to unified API (local and distributed)

vepadulano:
And also add tests that still use explicit full names to distributed RDataFrame functions and classes, to check compatibility with previous usage.

Sibling PR of https://github.com/root-project/root/pull/16681


martamaja10 (approved):
Looks good, thanks Vincenzo!

PR #1201, opened 2024-10-15
[hadd] add tests for hadd argument parsing

silverweed:
New test suite for hadd to go together with https://github.com/root-project/root/pull/16090


root/io/hadd/CMakeLists.txt
```diff
@@ -24,3 +25,105 @@ ROOTTEST_ADD_TEST(test_MergeNestedSameName
...
+# arg parsing tests
+# ------------------------------------
+
+ROOTTEST_ADD_TEST(test_hadd_args_gen_input
+                  COPY_TO_BUILDDIR hadd_gen_input_tree.C
+                  PRECMD ${CMAKE_COMMAND} -E rm -f hadd_args_in.root
+                  COMMAND ${ROOT_root_CMD} -q -b -l "hadd_gen_input_tree.C(\"hadd_args_in.root\")"
+)
+
+ROOTTEST_ADD_TEST(test_hadd_args_n0
```
pcanal, line 39: Does this test only verifies that the message printed by hadd are correct?  If yes, should we add some (minimal) test that the file produces is as expected (minimal requirement: contains the object we expects).

Rather than a `POSTCMD` deleting the file, I recommend a `PRECMD` (or not when we want to test that hadd complains about existing files).   We may also want to keep the file after the end to have an opportunity to hand check the file.

pcanal: Did we resolve this?

silverweed: @pcanal I just pushed a couple of new commits that should resolve it.

pcanal (approved):
Thanks.

PR #1203, opened 2024-10-16
adapt to new context menu function in th1

ferdymercury:
sibling of https://github.com/root-project/root/pull/16697


guitargeek (approved):
Thanks!

PR #1206, opened 2024-10-25
[cmake][win][skip-ci] Set the CMAKE_SKIP_TEST_ALL_DEPENDENCY variable

bellenot:
This should partially solve the issue with the rebuild (linking) of ROOT when building the tests.
See also: https://cmake.org/cmake/help/latest/variable/CMAKE_SKIP_TEST_ALL_DEPENDENCY.html
Requires CMake v3.29

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,7 @@ if(NOT DEFINED Python3_EXECUTABLE)
 endif()
 
 if(MSVC)
-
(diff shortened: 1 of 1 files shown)


pcanal:
Relates to https://github.com/root-project/root/pull/16744#issuecomment-2438231683

pcanal (approved):

PR #1213, opened 2024-11-07
[PyROOT] Skip numba test also if numba is not available at all

guitargeek:
So far, the test was only skipped if the numba version was too low, but it might also happen that numba is not installed at all. For example if one uses Python 3.13, which doesn't support numba yet.


vepadulano (approved):
Thanks!

PR #1216, opened 2024-11-13
remove complex-arg context menu functions

ferdymercury:
sibling of https://github.com/root-project/root/pull/16920


bellenot (approved):
LGTM

dpiparo:
Hi @ferdymercury unfortunately we have 1 failure on win32 due to the fact that the file `roottest/root/meta/MemberComments_win32.ref` was not updated. If you have time, could you please take care? If not, no problem, I can do that myself tomorrow.

PR #1218, opened 2024-11-14
add test for cv qualified vars

ferdymercury:
sibling of https://github.com/root-project/root/pull/16924


hahnjo:
Note that the commits should probably be squashed on merge

PR #1224, opened 2024-11-21
Adapt to change in error case behavior.

pcanal:
Companion PR of https://github.com/root-project/root/pull/16995

Instead of writing with the 'wrong' TStreamerInfo, the hadd of two file with conflicting StreamerInfo with the same version number write with the 'right' StreamerInfo ... but that StreamerInfo has missing fields with are now omitted in the writing.

The use case has 2 StreamerInfos:
(a)
v10 of ParticleImplVec
which inherits from v11 of Particle
which contains 3 floats

(b)
v10 (also) of ParticleImplVec
which inherits from v12 of Particle
which contains a float and moves the 3 pre-existing float into a base class (so they are 'missing' when using v11 of Particle).

The previous behavior of the hadd invocation was streaming
   v10-a of ParticleImplVec
but ended up continuing the streaming with
   v12 of Particle

Now the hadd invocation is still streaming
   v10-a of ParticleImplVec
but is continuing with streaming with its corresponding
   v11 of Particle
which has missing data member (those moved in the base) so during Streaming we could either write arbitrary values (0) or (the choice that is made) we just skip them leading to a more clear error at read time:

```
Error in <TBufferFile::CheckByteCount>: object of class vector<ParticleImplVec> read too many bytes: 56 instead of 8
```

Note that the `56` is actual another consequence of the 'broken' setup in which hadd had to run (duplicate version number) and the reading is incorrectly trying to read with version 12 of Particle.

Reminder: the point of these tests is to ensure that there is sufficient error output is case of this type of mixed up setup.

### DIFF
--- a/root/io/newstl/TestHelpers.C
+++ b/root/io/newstl/TestHelpers.C
@@ -6,7 +6,7 @@
 #include "TObjString.h"
 #include <utility>
 
-void fillListOfDir(TString directory, TList &l) {
+void fillListOfDir(const TString &directory, TList &l) {
    
    void *dir = gSystem->OpenDirectory(directory);
 
--- a/root/io/newstl/TestHelpers.C
+++ b/root/io/newstl/TestHelpers.C
@@ -6,9 +6,8 @@
 #include "TObjString.h"
 #include <utility>
 
-void fillListOfDir(TList &l) {
+void fillListOfDir(TString directory, TList &l) {
    
-   TString directory = ".";
    void *dir = gSystem->OpenDirectory(directory);
 
    const char *file = 0;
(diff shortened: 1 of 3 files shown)


root/io/newstl/TestHelpers.C
```diff
@@ -6,9 +6,8 @@
 #include "TObjString.h"
 #include <utility>
 
-void fillListOfDir(TList &l) {
+void fillListOfDir(TString directory, TList &l) {
```
dpiparo, line 9: const &?

dpiparo (approved):

root/io/newstl/TestHelpers.C
```diff
@@ -24,31 +23,45 @@ void fillListOfDir(TList &l) {
...
+   l.Sort();
 
+   if (gDebug > 3) {
       TIter next(&l);
       TObjString *obj;
       while ((obj = (TObjString*)next())) {
-         file = obj->GetName();
-//          cout << "found the directory " << obj->GetName() << endl;
+         const char *file = obj->GetName();
+         cout << "found the directory " << obj->GetName() << endl;
```
jblomer, line 60:
```suggestion
         const char *file = obj->GetName();
         cout << "found the directory " << file << endl;
```

root/io/newstl/TestHelpers.C
pcanal:
```suggestion
void fillListOfDir(const TString &directory, TList &l) {
```

PR #1229, opened 2024-11-30
[df] Add test with VariationsFor and no Vary

vepadulano:
The resulting map of variations should only have the nominal key.


dpiparo (approved):
lgtm, thanks.

PR #1230, opened 2024-12-02
[PyROOT] Don't use `force_flush` module

guitargeek:
With Python 3, force flushing is built into the `print()` function and enabled with an optional parameter.

This avoids failure in environments where the `force_flush` module is not available.


aaronj0 (approved):
LGTM!

PR #1234, opened 2024-12-03
Add test for distributed Alias

gpetruc:
from @vepadulano, to test https://github.com/root-project/root/pull/17171


vepadulano (approved):
Nice! Merging together with the sibling PR

PR #1235, opened 2024-12-03
Update tree tutorials path

mdessole:
This PR updates tree tutorials path within the ROOT library, according to https://github.com/root-project/root/pull/17097


guitargeek (approved):
Thanks!

PR #1237, opened 2024-12-03
Update execCheckNaming in windows

ferdymercury:
correct default allocator is now recognized

Before, it didn't recognize it was a default allocator since it was confused by the class / const.

Sibling of https://github.com/root-project/root/pull/17048


bellenot (approved):
LGTM, obviously after https://github.com/root-project/root/pull/17048 is merged

ferdymercury:
@pcanal @dpiparo sibling PR was merged, so this one should, too

PR #1247, opened 2025-01-02
io customizatipon rules: add test of changing the input type of sources.

pcanal:
And test of Array of objects and changing the underlying type

Companion to https://github.com/root-project/root/pull/17347


jblomer (approved):
Thanks, I find this test also a very useful reference.

PR #1252, opened 2025-01-25
io rules: test with array input followed by other inputs

pcanal:
This is the companion PR to https://github.com/root-project/root/pull/17523

### DIFF
--- a/root/io/evolution/rules/execSourceTypes.cxx
+++ b/root/io/evolution/rules/execSourceTypes.cxx
@@ -37,6 +37,8 @@ struct Old
    // and used in a read rule, the offset for the cached arrays was wrong - set to 0
    float fSingle = 1.0;
    int fInt = 1;
+   int fHitPattern[3] = {101, 102, 103};
+   int fHitCount = 2;
    A fValue{1.1};
    A fValueB{1.2};
    A fValueC{1.3};;
@@ -77,6 +79,8 @@ struct New
    double fSingle = 10.0;
    double fExtra  = 15.0;
    double fDouble = 20.0;
+   int fHitPattern[3] = {101, 102, 103};
+   int fHitCount = 2;
    B fValue;  // Using implicit conversion
    B fValueB; // Using explicit conversion from A to B
    C fValueC; // Using implicit conversion from A to B and explicit conversion to C
@@ -117,10 +121,16 @@ struct New
 
 #include <iostream>
 
+void examine(int a, bool trailing = true)
+{
+   std::cout << "The int value is: " << a;
+   if (trailing)
+      std::cout << '\n';
+}
+
 void examine(double a, bool trailing = true)
 {
-   std::cout << "The double";
-   std::cout << " value is: " << a;
+   std::cout << "The double value is: " << a;
    if (trailing)
       std::cout << '\n';
 }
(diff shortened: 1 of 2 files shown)


root/io/evolution/rules/execSourceTypes.cxx
```diff
@@ -237,13 +275,31 @@ bool check(T *p, V ref)
    return !error;
 }
 
+template<size_t N, typename V> //  = decltype(T::f)[N]>
+bool check_array(int arr[N], V ref)
```
jalopezg-git, line 279: I guess that we could make the two parameters of the same type (esp. given that they are being compared 1-to1 below), right?

root/io/evolution/rules/execSourceTypes.cxx
```diff
@@ -454,6 +514,9 @@ int readfile(const char *filename = "sourcetypes.root")
    res = res && check(n->fListExplicit, {"b1", "b2", "b3"});
    res = res && check(n->fListImplicitExplicit, {"c1", "c2", "c3"});
 
+   res = res && check_array<3>(n->fHitPattern, std::vector<int>{111, 112, 113});
```
jalopezg-git, line 517: Possibly use initializer list or `std::array<int, 3>` instead.
```suggestion
   res = res && check_array<3>(n->fHitPattern, std::initializer_list<int>{111, 112, 113});
```

root/io/evolution/rules/execSourceTypes.cxx
```diff
@@ -117,6 +121,14 @@ struct New
 
 #include <iostream>
 
+void examine(int a, bool trailing = true)
+{
+   std::cout << "The double";
+   std::cout << " value is: " << a;
```
jalopezg-git, line 127: Why `double`; isn't that and `int`?
```suggestion
   std::cout << "The integer value is: " << a;
```

jalopezg-git (commented):
Hey @pcanal; you invoked me (as reviewer), so here's the review (taking into account context from sibling root PR)!  Other than that, it LGTM!

Hope everything is going well at CERN / Fermilab - and Happy New 2025 :tada:!

dpiparo:
thanks @jalopezg-git !!!!!

jblomer (approved):

pcanal:
Thansk @jalopezg-git

root/io/evolution/rules/execSourceTypes.cxx
pcanal: Not quite.  `V` is meant to be any kind of collection (of int) with an `operator[]`

root/io/evolution/rules/execSourceTypes.cxx
pcanal: Indeed.  Thanks.

root/io/evolution/rules/execSourceTypes.cxx
pcanal: `std::initializer_list` does not worked because the type of the 2nd parameter is a template parameter of the function.  `std::array<int, 3>` could work but have the annoying need to repeat the `3` (actually I would love to not have it in the `check_array<3>` either but the naive tentatives to remove it did not work out.

PR #1259, opened 2025-02-12
Fix issues seen in high parallel ctest run

pcanal:
* root/io/hadd: Correct the filename needing to be checked (shown as reading a non-existent file)
* root/tree/evolution: Add missing dependency (shown as reading a partial build library file)

The companion PR is https://github.com/root-project/root/pull/17700


silverweed (approved):
So the hadd args tests were only doing half of their due checks...Thanks for fixing it.

PR #1261, opened 2025-02-13
Fix two test failures with functions marked nodiscard.

hageboeck:
With gcc-15, the dereferencing and the + operator of iterators are marked nodiscard. This makes two tests fail, because the warning leads to a non-zero diff.

### DIFF
--- a/cling/dict/runoperators.C
+++ b/cling/dict/runoperators.C
@@ -23,6 +23,6 @@ void runoperators() {
    std::vector<myiterator> v;
    std::vector<myiterator>::iterator iter;
    for( iter = v.begin(); iter != v.end(); ++iter) {
-      iter + 1;
+      static_cast<void>(iter + 1);
    }
 }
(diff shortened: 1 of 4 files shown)


cling/dict/runoperators.C
```diff
@@ -23,6 +23,6 @@ void runoperators() {
    std::vector<myiterator> v;
    std::vector<myiterator>::iterator iter;
    for( iter = v.begin(); iter != v.end(); ++iter) {
-      iter + 1;
+      static_cast<void>(= iter + 1);
```
pcanal, line 26:
```suggestion
      static_cast<void>(iter + 1);
```

pcanal (approved):
LGTM but we probably should have a companion PR in root to test the changes.

hageboeck:
> LGTM but we probably should have a companion PR in root to test the changes.

You are very right. It's running [here](https://github.com/root-project/root/pull/17719), and uncovered that I fixed two broken windows tests. 😅
```
TEST FAILURES:
1352:roottest-cling-dict-runoperators_WILL_FAIL
1451:roottest-cling-stl-default-VectorSort_WILL_FAIL
```
@bellenot

bellenot:
@hageboeck Cool! Congrats!

PR #1272, opened 2025-03-10
[df] Adjust test to debug warning

vepadulano:


root/dataframe/CMakeLists.txt
```diff
@@ -235,10 +235,18 @@ ROOTTEST_ADD_TEST(read_leaves_nodict
                   EXEC ./read_leaves_nodict
                   DEPENDS ${GENERATE_EXECUTABLE_TEST} read_leaves)
 
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
```
pcanal, line 238: This is technically a good test ... for now.  Namely this will fails the intent for a standalone `roottest` used to test an installed version of ROOT.   This will not longer be a problem when we merge the 2 repository .... in a few months.

root/dataframe/CMakeLists.txt
```diff
@@ -235,10 +235,18 @@ ROOTTEST_ADD_TEST(read_leaves_nodict
...
+# In debug mode RDataFrame warns when it needs to copy collection elements for reading
+ROOTTEST_ADD_TEST(test_readTotemNtuple
```
pcanal, line 240: To focus on the difference, I would make a variable to hold the reference filename rather than duplicating the full `ROOTTEST_ADD_TEST`.

pcanal: Another option (if the message issue during debug is not important) is to use an `OUTCNVCMD` to filter out the unwanted line.

pcanal (approved):

PR #1284, opened 2025-03-19
[CMake] Raise minimum required CMake version to 3.20

guitargeek:
To be consistent with the main ROOT repository and to avoid this warning:

```txt
CMake Deprecation Warning at roottest/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.9 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
 
 project(roottest)
 


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.9 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.20 FATAL_ERROR)
```
bellenot, line 1:
```suggestion
cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
```
To be consistent with ROOT

bellenot (approved):
But see my comment

CMakeLists.txt
guitargeek: Maybe you confused that with Cling? For ROOT the minimum is 3.20:
https://github.com/root-project/root/blob/master/CMakeLists.txt#L7

bellenot: Oh! OK, my bad! OK to go then

pcanal, line 1: What is the default CMake version on the oldest distribution we support?

guitargeek: It's 3.26.5 on `alma8`, so we're safe there. In any case, we don't have a choice I think because the minimum CMake version 3.20 is inherited from LLVM that we depend on as a builtin.

guitargeek:
I don't know why the CI failed on the main PR:

  * https://github.com/root-project/root/pull/18042

I will try to raise the minimum version to 3.10 only, let's see if it works then. Maybe the minimum cmake version has an effect on the CMake policies that are relevant for roottest?

guitargeek:
Seems to work now. We got to understand later why `roottest` doesn't work with 3.20 as the minimal required CMake version.

PR #1301, opened 2025-04-11
[io] Do not run prefetching tests if xrootd is not available

dpiparo:


dpiparo:
fixes https://github.com/root-project/root/issues/17454

PR #1302, opened 2025-04-11
Test for name normalization involving size_t template argument

pcanal:
This is the companion PR of https://github.com/root-project/root/pull/18373

### DIFF
--- a/root/meta/naming/issue-18363/Objects.h
+++ b/root/meta/naming/issue-18363/Objects.h
@@ -3,13 +3,10 @@
 
 namespace reco { 
 template <size_t I, bool B> struct PFRecHitSoALayout {
-  template <size_t II, bool B1, bool B2, bool B3> struct ViewTemplateFreeParams {};
-  using View = ViewTemplateFreeParams<I, B, true, true>; 
-}; 
-using PFRecHitSoA = PFRecHitSoALayout<128UL,false>; // try with 128ul
+   template <size_t II, bool B1, bool B2, bool B3> struct ViewTemplateFreeParams {};
+   using View = ViewTemplateFreeParams<I, B, true, true>;
+};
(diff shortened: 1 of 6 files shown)


root/meta/naming/issue-18363/Objects.h
```diff
@@ -0,0 +1,12 @@
+#ifndef Objects_18363_hxx
+#define Objects_18363_hxx
+
+namespace reco { 
+template <size_t I, bool B> struct PFRecHitSoALayout {
+   template <size_t II, bool B1, bool B2, bool B3> struct ViewTemplateFreeParams {};
+   using View = ViewTemplateFreeParams<I, B, true, true>;
+};
+using PFRecHitSoA = PFRecHitSoALayout<128,false>; // The result is the same with 128UL or 128ul
+}
```
hahnjo, line 10: The minimal reproducer to trigger the problem is:
```
template <unsigned I> struct PFRecHitSoALayout {
   struct View {}; 
};
```
This normalizes to `PFRecHitSoALayout<128U>::View`.

A couple of conclusions:
1. The problem is not particular to `size_t`, it just has to be a builtin type where `printIntegral` decides to print a suffix.
2. It has nothing to do with `using` declarations that "forward" a template argument.
3. It is not even required for the inner type to be a `template`.

FYI @vgvassilev (not a standalone reproducer yet)

pcanal: If I recalled correctly, it does not add the suffix for int.  I.e. being an `unsigned` type might be the discriminant.

hahnjo: Yes, as I write
> it just has to be a builtin type where `printIntegral` decides to print a suffix

That's not the case for `int`, but also includes `signed long` for example.

pcanal:
This was replaced by https://github.com/root-project/root/pull/18464

PR #1303, opened 2025-04-14
[io] Fix unsequenced operation in IoBigEventGeneration

silverweed:
The order of operations for fNevent was unspecified. Splitting the two expressions to make it well-defined.


guitargeek (approved):
Thanks!

hahnjo (approved):

PR #1304, opened 2025-04-14
[io] remove MathCore dependency from IoBigEventGeneration

silverweed:
It's not needed and it makes CMake fail to configure


guitargeek (approved):
Thanks! I see how this fixes the `roottest` build, because `MathCore` comes from ROOT. You're probably building `roottest` standalone, no?

guitargeek:
Not relevant anymore since `roottest` got merged into `root`.

guitargeek:
Actually, let's maybe merge this anyway since people in the forum are talking about it:
https://root-forum.cern.ch/t/cea-45230-roottest-v6-36-00-iobigeventgenerator-mathcore-dependency-issue/63654

guitargeek:
Can't merge because branch is locked.

PR #1329, opened 2025-06-05
Fix pragma

linev:
Use `#pragma link C++ class vector<Track>+` instead of `#pragma link C++ vector<Track>+`


pcanal (approved):
Thanks.

hahnjo:
This should not have been merged, the `master` branch is effectively closed after the integration into the main `root.git`!

linev:
@hahnjo 
Thanks, I see it now.
I apply fix to 6.36 branch where it still used.

